PR #3, opened 2022-06-30
Update go module, and copyright

c3y1huang:


c3y1huang:
I cannot add reviewers.
PTAL @longhorn/maintainer

shuo-wu (approved):

go.mod
```diff
@@ -1,14 +1,8 @@
-module github.com/keithalucas/longhorn-spdk
+module github.com/longhorn/longhorn-spdk
 
 go 1.13
```
joshimoo, line 3: Let's update the go version to v1.17 same as in the manager.
This also changes the way modules are handled a bit.

joshimoo (commented):
@innobead @yasker can you update the repositories permissions, doesn't seem like maintainers can approve/request reviewers.

Do we want to keep the copyright notices per file? Seems redundant (since all of it is MIT licensed) and different from the other longhorn repositories.



Let's update the go version to 1.17 to be in sync with the other longhorn repositories, this also changes the module handling a bit.

go.mod
c3y1huang: Updated.

c3y1huang:
> Do we want to keep the copyright notices per file? Seems redundant (since all of it is MIT licensed) and different from the other longhorn repositories.

Removed.

innobead (approved):

innobead:
Thanks @c3y1huang 👍

PR #4, opened 2023-05-11
New SPDK engine implementation

shuo-wu:
- Removed the deprecated implementation, which is mainly covered by `longhorn/go-spdk-helper` now.
- Introduced a SPDK gRPC server for engine and replica handling.


shuo-wu:
Let merge this then Derek can handle the disk part

PR #6, opened 2023-05-11
Fix wrong returned values and missing apis

derekbit:


proto/ptypes/spdk.proto
```diff
@@ -9,14 +9,14 @@ service SPDKService {
     rpc ReplicaDelete(ReplicaDeleteRequest) returns (google.protobuf.Empty);
     rpc ReplicaGet(ReplicaGetRequest) returns (Replica);
     rpc ReplicaSnapshotCreate(SnapshotRequest) returns (Replica);
-    rpc ReplicaSnapshotDelete(SnapshotRequest) returns (google.protobuf.Empty);
```
shuo-wu, line 12: For the deletion, there is no need to or we may be not able to return the object...
Same below

derekbit: Got it. Return `&empty. Empty{}` instead. Thank you.

innobead (approved):

PR #7, opened 2023-05-11
Rename ptypes to spdkrpc

derekbit:
ptypes is used by longhorn-engine's syncagent grpc service


shuo-wu:
`ptypes` means proto types. Do we need to have a different naming for each repo?

derekbit:
> `ptypes` means proto types. Do we need to have a different naming for each repo?

Yes, we need a different for avoiding the proto.Register confilicts of legacy engine and spdk engine.

shuo-wu:
Got it! I almost forgot that proto issue...

shuo-wu (approved):

innobead:
`ptypes` pkg name is too general in longhorn-engine. For any gRPC naming should be more specific in the future.

PR #8, opened 2023-05-11
Update replica

shuo-wu:


innobead:
@shuo-wu need to fix the conflicts.

proto/ptypes/spdk.proto
```diff
@@ -37,8 +37,11 @@ message Replica {
     string lvs_uuid = 4;
     uint64 spec_size = 5;
     uint64 actual_size = 6;
-    map<string, Lvol> snapshots = 7;
-    bool rebuilding = 8;
+    string ip = 7;
+    int32 first_port = 8;
+    int32 last_port = 9;
```
derekbit, line 42: Can we use `port_start` and `port_end`?

shuo-wu: Updated

derekbit (approved):

derekbit (dismissed):

shuo-wu:
Do you know how to fix the error `Complex Code` reported by `CodeFactor`?

innobead:
Let me remove this check first since this should not be enforced check.

shuo-wu:
@innobead Would you help disable the failed checks? Thank you!

innobead:
> @innobead Would you help disable the failed checks? Thank you!

It was weird since it has been removed from the repo setting but still existed. It needs to be removed from the code refractory setting page instead. It is removed for now. Let's add it back after 1.5.

pkg/spdk/server.go
```diff
@@ -71,59 +71,83 @@ func (s *Server) monitoringReplicas() {
...
 	r := s.replicaMap[req.Name]
-	s.Unlock()
 
-	portStart, _, err := s.portAllocator.AllocateRange(1)
-	if err != nil {
-		return nil, err
-	}
-
-	return r.Create(s.spdkClient, portStart, req.ExposeRequired)
+	return r.Create(s.spdkClient, req.ExposeRequired, s.portAllocator)
```
derekbit, line 135: Remove `r` from `replicaMap` if it is unable to be created.

pkg/spdk/server.go
```diff
@@ -71,59 +71,83 @@ func (s *Server) monitoringReplicas() {
 			logrus.Info("SPDK Server: stopped monitoring replicas due to the context done")
 			done = true
 		case <-ticker.C:
-			bdevLvolList, err := s.spdkClient.BdevLvolGet("", 0)
-			if err != nil {
-				logrus.Errorf("SPDK Server: failed to get lvol bdevs for replica cache update, will retry later: %v", err)
-				continue
+			if err := s.verifyReplicas; err != nil {
+				logrus.Errorf("SPDK Server: failed to verify and update replica cache, will retry later: %v", err)
```
derekbit, line 75: nit:
```suggestion
				logrus.WithError(err).Errorf("SPDK Server: failed to verify and update replica cache, will retry later")
```

pkg/spdk/server.go
```diff
@@ -71,59 +71,83 @@ func (s *Server) monitoringReplicas() {
...
+	}
+
+	return nil
+}
+
 func (s *Server) ReplicaCreate(ctx context.Context, req *spdkrpc.ReplicaCreateRequest) (ret *spdkrpc.Replica, err error) {
+	if req.Name == "" {
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "replica name is required")
+	}
+	if req.LvsName == "" || req.LvsUuid == "" {
```
derekbit, line 121: Do we really need `LvsName`?

pkg/spdk/server.go
```diff
@@ -71,59 +71,83 @@ func (s *Server) monitoringReplicas() {
 			logrus.Info("SPDK Server: stopped monitoring replicas due to the context done")
 			done = true
 		case <-ticker.C:
-			bdevLvolList, err := s.spdkClient.BdevLvolGet("", 0)
-			if err != nil {
-				logrus.Errorf("SPDK Server: failed to get lvol bdevs for replica cache update, will retry later: %v", err)
-				continue
+			if err := s.verifyReplicas; err != nil {
```
derekbit, line 74: What's the purpose of the verification?

pkg/spdk/server.go
shuo-wu: The current strategy is that retaining invalid/error replicas then the caller is responsible for cleaning them up

pkg/spdk/server.go
shuo-wu: Do you mean the input? We need to retain lvsName in cache `Replica` anyway (for local replica identification during engine creation). For the input, I am fine with removing it. But we need to get the lvs from SPDK if the name not provided.

pkg/spdk/server.go
derekbit: I see. We can keep it here.

pkg/spdk/server.go
derekbit: I see. We can keep it here.

pkg/spdk/server.go
derekbit: Got it. If the creation fails, I prefer to clean it up before exiting the function, but I'm fine with the implementation.

derekbit: Another question, does legacy volume replica also adopt the same strategy? If no, will we need a different error handling logic in longhorn-manager?

pkg/spdk/replica.go
```diff
@@ -218,7 +229,40 @@ func (r *Replica) ValidateAndUpdate(bdevLvolMap map[string]*spdktypes.BdevInfo)
...
+			continue
+		}
+		exposedPort, err = strconv.Atoi(listenAddr.Trsvcid)
+		if err != nil {
+			return err
+		}
+		break
+	}
+	if int32(exposedPort) != r.PortStart {
+		if stopErr := spdkClient.StopExposeBdev(nqn); stopErr != nil {
```
derekbit, line 261: Do we need a lock protection for a spdkClient operation?

derekbit: I was thinking the same question before, but I am fine with it fine for now.

shuo-wu: Sure. I will add it after this PR merge

derekbit (approved):
LGTM

derekbit (approved):
LGTM

derekbit (approved):
LGTM

pkg/spdk/replica.go
```diff
@@ -372,39 +380,55 @@ func (r *Replica) Create(spdkClient *spdkclient.Client, port int32, exposeRequir
...
 		}
-		if exposeStopped {
-			r.IP = ""
-			r.Port = 0
+		// The port can be released once the rebuilding and expose are stopped
+		if !r.IsExposed && r.PortStart != 0 {
+			if releaseErr := superiorPortAllocator.ReleaseRange(r.PortStart, r.PortEnd); releaseErr != nil {
+				r.log.Errorf("Failed to release port %d to %d at the end of replica deletion: %v", r.PortStart, r.PortEnd, releaseErr)
+				return
+			}
```
innobead, line 454: In this case, it will be orphaned replica? how can these ports be reused eventually?

innobead (approved):
In general, LGTM.

pkg/spdk/replica.go
shuo-wu: Ideally, the monitor will periodically list all lvols and build cache for each RW lvol, which is also the missing part for the disk create/add API.
I will handle it after the demo.

PR #9, opened 2023-05-12
Introduce disk methods

derekbit:
Made a mistake. The `disk.go` (server side implementation) is already added in the [previous PR](https://github.com/longhorn/longhorn-spdk-engine/pull/7/files#diff-1072a078f4eaf6f8e907ac70d01de1665ad0dbe339f4bda39006a7e388a298af).


proto/spdkrpc/spdk.proto
```diff
@@ -18,6 +18,10 @@ service SPDKService {
     rpc EngineSnapshotCreate(SnapshotRequest) returns (Engine);
     rpc EngineSnapshotDelete(SnapshotRequest) returns (google.protobuf.Empty);
 
+    rpc DiskCreate(DiskCreateRequest) returns (Disk);
+    rpc DiskDelete(DiskDeleteRequest) returns (google.protobuf.Empty);
+	rpc DiskGet(DiskGetRequest) returns (Disk);
```
innobead, line 23: fix the wrong indents.

pkg/client/client.go
```diff
@@ -174,3 +174,50 @@ func (c *SPDKClient) EngineGet(name string) (*api.Engine, error) {
...
+
+	return client.DiskCreate(ctx, &spdkrpc.DiskCreateRequest{
+		DiskName:  diskName,
+		DiskPath:  diskPath,
+		BlockSize: blockSize,
+	})
+}
+
+func (c *SPDKClient) DiskGet(diskName, diskPath string) (*spdkrpc.Disk, error) {
+	if diskPath == "" {
```
innobead, line 193: Do we need to validate `diskName` as well?

pkg/spdk/disk.go
```diff
@@ -122,7 +122,7 @@ func (s *Server) DiskGet(ctx context.Context, req *spdkrpc.DiskGetRequest) (*spd
 	defer s.RUnlock()
 
 	// Check if the disk exists
-	bdevs, err := s.spdkClient.BdevAioGet(req.DiskName, 3000)
+	bdevs, err := s.spdkClient.BdevAioGet(req.DiskName, 0)
```
innobead, line 125: What's the reason for better response time?

pkg/spdk/disk.go
```diff
@@ -159,13 +159,13 @@ func getDiskPath(path string) string {
...
 func getDeviceNumber(filename string) (string, error) {
-	fi, err := os.Stat(filename)
+	fileInfo, err := os.Stat(filename)
 	if err != nil {
 		return "", err
 	}
-	dev := fi.Sys().(*syscall.Stat_t).Dev
-	major := int(dev >> 8)
-	minor := int(dev & 0xff)
+	dev := fileInfo.Sys().(*syscall.Stat_t).Rdev
```
innobead, line 166: Can you explain the purpose of this change?

innobead (changes requested):
In general, LGTM. Just a few comments.

pkg/spdk/disk.go
derekbit: The purpose of the timeout is waiting for a aio bdev with a given name to appear. For `DiskGet`, it is not necessary for the wait. If the bdev does not exist, it means the bdev is not created.

pkg/client/client.go
derekbit: Yeah, forgot addint it.
Added, thank you.

proto/spdkrpc/spdk.proto
derekbit: Fixed. Thank you.

pkg/spdk/disk.go
derekbit: `sdev` is the ID of device containing file.
`rdev` is the device ID.

The function is getting the device ID (major-minor) of the give disk path.

innobead (approved):

pkg/spdk/disk.go
```diff
@@ -159,13 +159,13 @@ func getDiskPath(path string) string {
 }
 
 func getDeviceNumber(filename string) (string, error) {
```
shuo-wu, line 161: Can we use `DetectDevice` in `go-spdk-helper/pkg/util/device.go`?

shuo-wu (commented):
1. Is it better to move the API functions from `spdk/disk.go` to `spdk/server.go`? I mean, the APIs like `func (s *Server) DiskCreate(ctx context.Context, req *spdkrpc.DiskCreateRequest) (*spdkrpc.Disk, error)`
2. Why not use `spdkClient.DeleteDevice()` in API `DiskDelete`?

derekbit:
> 2\. Why not use `spdkClient.DeleteDevice()` in API `DiskDelete`?

Didn't notice the api.
`spdkClient.DeleteDevice() ` deletes the lvstore. Do we need to delete it?

pkg/spdk/disk.go
derekbit: Yes, same result. will use it instead.

shuo-wu:
OK. Nice catch! It seems we cannot use it... But do we need to provide an option/API to help users erase the lvol data (using `DeleteDevice`) rather than simply detach the disk then?

derekbit:
> OK. Nice catch! It seems we cannot use it... But do we need to provide an option/API to help users erase the lvol data (using `DeleteDevice`) rather than simply detach the disk then?

No. Deleting a disk just removes the disk from longhorn-system by removing it from `node.disks` as the previous behavior. I'm thinking how to send the `erase` signal to spdk service.

innobead:
What's the current behavior of removing disk in the current stack?

derekbit:
> What's the current behavior of removing disk in the current stack?

Just remove the `disk` from `node. Disks` and don't erase disk.

innobead:
If unable to achieve it now in SPDK, let's just create an issue to improve later. Probably together with the old stack.

derekbit:
> If unable to achieve it now in SPDK, let's just create an issue to improve later. Probably together with the old stack.

Tracked in the ticket https://github.com/longhorn/longhorn/issues/5925.

PR #11, opened 2023-05-12
Get device ID using DetectDevice()

derekbit:


pkg/spdk/server.go
```diff
@@ -18,6 +22,11 @@ import (
 
 const (
 	MonitorInterval = 3 * time.Second
+
+	defaultClusterSize = 4 * 1024 * 1024 // 4MB
```
shuo-wu, line 26: Actually, I am using 1MiB as the default value for API `spdkClient.DeviceAdd`

pkg/spdk/disk.go
```diff
@@ -10,163 +9,26 @@ import (
...
-}
-
 func getDiskPath(path string) string {
 	return filepath.Join(hostPrefix, path)
 }
 
-func getDeviceNumber(filename string) (string, error) {
-	fileInfo, err := os.Stat(filename)
+func getDiskID(filename string) (string, error) {
+	executor := util.NewTimeoutExecutor(util.CmdTimeout)
```
innobead, line 25: nit: `util.CmdTimeout` is 1 min, but getting disk ID should be quick enough, so probably just use a no-timeout executor for this kind of operations.

innobead (approved):
LGTM.

pkg/spdk/server.go
```diff
@@ -206,6 +215,130 @@ func (s *Server) EngineSnapshotDelete(ctx context.Context, req *spdkrpc.Snapshot
...
+	log.Info("Creating disk")
+
+	if req.DiskName == "" || req.DiskPath == "" {
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "disk name and disk path are required")
+	}
+
+	s.Lock()
+	defer s.Unlock()
+
+	if err := s.validateDiskCreateRequest(req); err != nil {
```
shuo-wu, line 234: This can be a single function rather than a `Server` method.

pkg/spdk/server.go
```diff
@@ -206,6 +215,130 @@ func (s *Server) EngineSnapshotDelete(ctx context.Context, req *spdkrpc.Snapshot
...
+		blockSize = uint64(req.BlockSize)
+	}
+
+	uuid, err := s.addBlockDevice(req.DiskName, req.DiskPath, blockSize)
+	if err != nil {
+		log.WithError(err).Error("Failed to add block device")
+		return nil, grpcstatus.Error(grpccodes.Internal, errors.Wrap(err, "failed to add block device").Error())
+	}
+
+	return s.lvstoreToDisk(req.DiskPath, "", uuid)
```
shuo-wu, line 251: Similarly, this can be a single function rather than a `Server` method.

pkg/spdk/server.go
```diff
@@ -206,6 +215,130 @@ func (s *Server) EngineSnapshotDelete(ctx context.Context, req *spdkrpc.Snapshot
...
+	})
+
+	log.Info("Creating disk")
+
+	if req.DiskName == "" || req.DiskPath == "" {
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "disk name and disk path are required")
+	}
+
+	s.Lock()
+	defer s.Unlock()
```
shuo-wu, line 232: Do we need this lock? I am thinking if this lock is necessary for using `s.spdkClient`

pkg/spdk/server.go
```diff
@@ -206,6 +215,130 @@ func (s *Server) EngineSnapshotDelete(ctx context.Context, req *spdkrpc.Snapshot
...
+
+	lvstores, err := s.spdkClient.BdevLvolGetLvstore("", req.DiskUuid)
+	if err != nil {
+		resp, parseErr := parseErrorMessage(err.Error())
+		if parseErr != nil || !isNoSuchDevice(resp.Message) {
+			return nil, errors.Wrapf(err, "failed to get lvstore with UUID %v", req.DiskUuid)
+		}
+		log.WithError(err).Errorf("Cannot find lvstore with UUID %v", req.DiskUuid)
+		return nil, grpcstatus.Errorf(grpccodes.NotFound, "cannot find lvstore with UUID %v", req.DiskUuid)
```
shuo-wu, line 276: Can we continue deleting the aio bdev when lvstore is not found?

pkg/spdk/server.go
derekbit: Not really sure if the spdk_tgt respond in order...

pkg/spdk/server.go
derekbit: Sure.

PR #12, opened 2023-05-12
Implement replica and engine watch methods

derekbit:


derekbit:
This is an ad-hoc impl. @shuo-wu will take over it.

PR #13, opened 2023-05-13
Make server.go simple and move the implementations into disk.go

derekbit:


pkg/spdk/server.go
```diff
@@ -22,7 +19,7 @@ import (
 const (
 	MonitorInterval = 3 * time.Second
 
-	defaultClusterSize = 4 * 1024 * 1024 // 4MB
+	defaultClusterSize = 1 * 1024 * 1024 // 1MB
 	defaultBlockSize   = 4096            // 4KB
```
shuo-wu, line 23: NIT: Move these to `disk.go`

shuo-wu (approved):
In general LGTM

pkg/spdk/server.go

shuo-wu (approved):

PR #15, opened 2023-05-14
Fix wrong unlock

derekbit:


shuo-wu (approved):

pkg/spdk/server.go
```diff
@@ -118,7 +118,7 @@ func (s *Server) ReplicaDelete(ctx context.Context, req *spdkrpc.ReplicaDeleteRe
 	s.RLock()
 	r := s.replicaMap[req.Name]
 	delete(s.replicaMap, req.Name)
-	s.Unlock()
+	s.RUnlock()
```
shuo-wu, line 121: Sorry for missing some important issues before merge:
1. If you directly delete a replica from the map here. This should not be a read lock.
2. Can we directly remove replicas from the map without considering whether the below cleanup succeeds?
3. One thing I forgot to update is: we will remove replicas from the map only when `req.CleanupRequired` is true

derekbit: > Can we directly remove replicas from the map without considering whether the below cleanup succeeds?

I think no. In addition, there are many parts needing cleanup logics in the engine implementation.

> One thing I forgot to update is: we will remove replicas from the map only when req.CleanupRequired is true

Got it. Will you handle this part?

shuo-wu: I already add the fix commits in https://github.com/longhorn/longhorn-spdk-engine/pull/8

shuo-wu: Updated the PR with all mentioned issues fixed

PR #18, opened 2023-05-15
Use bdev names instead of lvol names for creating a raid

derekbit:


pkg/spdk/engine.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/longhorn/longhorn-spdk-engine/proto/spdkrpc"
 )
 
-func SvcEngineCreate(spdkClient *spdkclient.Client, name, frontend string, replicaAddressMap map[string]string, port int32) (ret *spdkrpc.Engine, err error) {
+func SvcEngineCreate(spdkClient *spdkclient.Client, name, frontend string, bdevAddressMap map[string]string, port int32) (ret *spdkrpc.Engine, err error) {
 	if frontend != types.FrontendSPDKTCPBlockdev && frontend != types.FrontendSPDKTCPNvmf {
```
innobead, line 24: Can we make this function private unexported? otherwise, need to do`replicaAddressMapToBdevAddressMap` in this function instead of callers to prevent redundant calls to conver replica addresses to bdev addresses.

innobead (changes requested):
LGTM, but just one comment needed to resolve.

pkg/spdk/engine.go
derekbit: Do you mean `func (s *server) svcEngineCreate(...)`?

derekbit: I prefer it as well.

innobead (approved):

PR #20, opened 2023-05-15
Fix ReplicaModeMap

derekbit:
keys of ReplicaModeMap are addresses.

### DIFF
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -205,8 +205,9 @@ func svcEngineGet(spdkClient *spdkclient.Client, name string) (res *spdkrpc.Engi
 		if !exists {
 			// This replica must be a local lvol
 			replicaName := spdktypes.GetLvolNameFromAlias(baseBdev.Name)
-			res.ReplicaAddressMap[replicaName] = ""
-			res.ReplicaModeMap[replicaName] = spdkrpc.ReplicaMode_RW
+			address := fmt.Sprintf("%s:0", podIP)
+			res.ReplicaAddressMap[replicaName] = address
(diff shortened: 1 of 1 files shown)


innobead (approved):

shuo-wu (commented):
Do we need to modify the input address map here? I am thinking if we can ask the replica to return the correct address

derekbit:
> Do we need to modify the input address map here? I am thinking if we can ask the replica to return the correct address

The `refresh` function in the engine controller triggers a call to `ReplicaList` in order to obtain the replica list. In the case of SPDK, this call to `ReplicaList` actually triggers `EngineGet`, which returns both the `ReplicaAddressMap` and `ReplicaModeMap`. Since the `ReplicaAddressMap` already provides the mapping of replica name and address, it can be used to obtain the mapping of address to mode. This enables the retrieval of all necessary information.

shuo-wu:
I mean, when creating a engine, the input `replicaAddressMap` should contain valid addresses (the port may be 0 bu the format is correct) regardless of local replica exists or not. Then we don't need to do anything extra for engine create or cache update.

derekbit:
> I mean, when creating a engine, the input `replicaAddressMap` should contain valid addresses (the port may be 0 bu the format is correct) regardless of local replica exists or not. Then we don't need to do anything extra for engine create or cache update.

Oh! Sorry for misunderstanding the question.
I think the input address map should be valid and don't think need to get the address from replicas again.
We can improve it in the next phase, WDYT?

shuo-wu:
Can we hold this PR and revisit it after merging the engine branch I mentioned? This makes that rebase messy anyway

derekbit:
After discussing with @shuo-wu , we can keep the current implementation and don't need this PR.

PR #24, opened 2023-05-16
Set local replica address to podIP:0

derekbit:


pkg/spdk/engine.go
```diff
@@ -205,7 +205,7 @@ func svcEngineGet(spdkClient *spdkclient.Client, name string) (res *spdkrpc.Engi
 		if !exists {
 			// This replica must be a local lvol
 			replicaName := spdktypes.GetLvolNameFromAlias(baseBdev.Name)
-			res.ReplicaAddressMap[replicaName] = ""
+			res.ReplicaAddressMap[replicaName] = fmt.Sprintf("%s:0", podIP)
```
shuo-wu, line 208: NIT: use `net.JoinHostPort()`

derekbit: One question:
Does local replica still allocate a port? @shuo-wu

derekbit:
No need after introducing engine cache.

PR #31, opened 2023-05-18
Introduce field Engine.State and Implement watch APIs

shuo-wu:


pkg/spdk/replica.go
```diff
@@ -377,16 +387,33 @@ func constructSnapshotMap(replicaName string, rootSvcLvol *Lvol, bdevLvolMap map
...
-	if r.State != ReplicaStatePending && r.State != ReplicaStateStopped {
-		return nil, fmt.Errorf("invalid state %s for replica %s creation", r.Name, r.State)
+	if r.State == types.InstanceStateRunning {
+		updateRequired = false
+		return nil, grpcstatus.Errorf(grpccodes.AlreadyExists, "replica %v already exists and running", r.Name)
+	}
+	if r.State == types.InstanceStateStopped && !exposeRequired {
+		updateRequired = false
+		return nil, grpcstatus.Errorf(grpccodes.AlreadyExists, "replica %v already exists without expose", r.Name)
+	}
```
derekbit, line 408: No need to error out here.
It is for the local lvol case. Still need to return replica, set it to running and update.

derekbit:
BTW, can you help update the two functions in `pkg/api/types.go`? Thank you.

```
func (s *ReplicaStream) Recv() (*empty.Empty, error) {
	return s.stream.Recv()
}

func (s *EngineStream) Recv() (*empty.Empty, error) {
	return s.stream.Recv()
}
```

pkg/spdk/replica.go
shuo-wu: Here I would like to make the behavior consistent:
1. If the replica is running/exposed, trying to expose it again would error out.
2. If the local replica is created but does not need to be exposed, trying to create it without expose again would error out as well.

shuo-wu:
> BTW, can you help update the two functions in pkg/api/types.go? Thank you.

Function updated.

pkg/spdk/replica.go
derekbit: For the a running lvol, the previous check `if r.State == types.InstanceStateRunning { .. }` can error out.
 
But for the lvol, we still transient the state from `stopped `to `running`.

shuo-wu: After discussing withe Derek, we will remove this check. 
`InstanceStateStopped` means the replica releases ports/hasn't got ports. And it's valid for API `Create`

PR #38, opened 2023-05-23
Add API EngineReplicaDelete & Verify Engine map consistency 

shuo-wu:


pkg/spdk/engine.go
```diff
@@ -458,6 +458,39 @@ func (e *Engine) validateAndUpdateReplicaMode(replicaName string, bdev *spdktype
...
+		return fmt.Errorf("cannot find replica name with address %s for engine %s replica delete", replicaAddress, e.Name)
+	}
+	if e.ReplicaAddressMap[replicaName] == "" {
+		return fmt.Errorf("replica %s is not found for engine %s replica delete", replicaName, e.Name)
+	}
+	if replicaAddress != "" && e.ReplicaAddressMap[replicaName] != replicaAddress {
+		return fmt.Errorf("replica %s recorded address %s does not match the input address %s for engine %s replica delete", replicaName, e.ReplicaAddressMap[replicaName], replicaAddress, e.Name)
+	}
+
+	if _, err := spdkClient.BdevRaidRemoveBaseBdev(e.ReplicaBdevNameMap[replicaName]); err != nil {
```
shuo-wu, line 518: Not sure if we need to mark the replica mode as ERR if this fails

shuo-wu: If the SPDK server actually does not do anything for the base bdev, YES the base bdev should still function.
If the SPDK server tried to operate the base bdev but failed in the middle, then we would assume the SPDK server will mark the base bdev as unavailable by itself. Longhorn can blindly sync-up with it.

derekbit (approved):
LGTM

PR #43, opened 2023-05-29
Remove diskPath from DiskGet

derekbit:


pkg/util/util.go
```diff
@@ -29,3 +30,10 @@ func GetVolumeNameFromEngineName(engineName string) string {
 func BytesToMiB(bytes uint64) uint64 {
 	return bytes / 1024 / 1024
 }
+
+func RemovePrefix(path, prefix string) string {
+	if strings.HasPrefix(path, prefix) {
+		return strings.TrimPrefix(path, prefix)
+	}
+	return path
```
innobead, line 38: Can just `return strings.TrimPrefix(path, prefix)`

innobead (approved):
LGTM.

pkg/util/util.go
shuo-wu: NIT: is this function necessary...? Why not directly call `strings.HasPrefix`...?

derekbit: Not necessary. Can do `diskPath := strings.TrimPrefix(path, prefix)` directly

PR #45, opened 2023-06-05
Offline rebuilding

shuo-wu:


proto/spdkrpc/spdk.proto
```diff
@@ -21,6 +27,7 @@ service SPDKService {
     rpc EngineSnapshotDelete(SnapshotRequest) returns (google.protobuf.Empty);
     rpc EngineList(google.protobuf.Empty) returns (EngineListResponse);
     rpc EngineWatch(google.protobuf.Empty) returns (stream google.protobuf.Empty) {}
+    rpc EngineReplicaAdd(EngineReplicaAddRequest) returns (google.protobuf.Empty) {}
     rpc EngineReplicaDelete(EngineReplicaDeleteRequest) returns (google.protobuf.Empty) {}
 
     rpc DiskCreate(DiskCreateRequest) returns (Disk);
```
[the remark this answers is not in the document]
shuo-wu: The legacy rebuilding requires extra orchestration/handling in the proxy server `ReplicaAdd`. I mean, creating a `task` to handle the rebuilding flow. But for SPDK rebuilding, it just needs to call this gRPC API.

shuo-wu:
Manual Test:
- [x] Replica snapshot APIs
- [x] Engine snapshot APIs
- [ ] Engine ReplicaAdd (offline rebuilding) API

pkg/spdk/engine.go
```diff
@@ -180,6 +156,38 @@ func getBdevNameForReplica(spdkClient *spdkclient.Client, localReplicaBdevMap ma
...
+		return fmt.Errorf("unknown frontend type %s", e.Frontend)
+	}
+
+	if e.Frontend == types.FrontendSPDKVoid {
+		e.log.Infof("No frontend specified, will not expose the volume %s", e.VolumeName)
+		return nil
+	}
+
+	nqn := helpertypes.GetNQN(e.Name)
+	port, _, err := superiorPortAllocator.AllocateRange(1)
```
derekbit, line 170: nit: make 1 as a const `DefaultEngineReservedPortCount`.

pkg/spdk/engine.go
```diff
@@ -180,6 +156,38 @@ func getBdevNameForReplica(spdkClient *spdkclient.Client, localReplicaBdevMap ma
...
+	if err != nil {
+		return err
+	}
+	if err := spdkClient.StartExposeBdev(nqn, e.Name, e.IP, strconv.Itoa(int(port))); err != nil {
+		return err
+	}
+	e.Port = port
+
+	if e.Frontend == types.FrontendSPDKTCPNvmf {
+		e.Endpoint = GetNvmfEndpoint(nqn, e.IP, e.Port)
```
derekbit, line 180: The format `fmt.Sprintf("%s://%s:%d", nqn, ip, port)` looks a bit weird. Is it correct?

pkg/spdk/engine.go
```diff
@@ -122,51 +123,26 @@ func (e *Engine) Create(spdkClient *spdkclient.Client, replicaAddressMap, localR
...
+func getBdevNameForReplica(spdkClient *spdkclient.Client, localReplicaLvsNameMap map[string]string, replicaName, replicaAddress, podIP string) (bdevName string, err error) {
 	replicaIP, replicaPort, err := net.SplitHostPort(replicaAddress)
 	if err != nil {
 		return "", err
 	}
 	if replicaIP == podIP {
-		if localReplicaBdevMap[replicaName] == "" {
-			return "", fmt.Errorf("cannot to find local replica %s from the local replica bdev map", replicaName)
+		if localReplicaLvsNameMap[replicaName] == "" {
+			return "", fmt.Errorf("cannot to find local replica %s from the local replica map", replicaName)
```
derekbit, line 142: "failed to ..."

pkg/spdk/engine.go
```diff
@@ -498,6 +528,223 @@ func (e *Engine) validateAndUpdateReplicaMode(replicaName string, bdev *spdktype
...
+	e.Lock()
+	defer e.Unlock()
+
+	// Syncing with the SPDK TGT server only when the engine is running.
+	if e.State != types.InstanceStateRunning {
+		return fmt.Errorf("invalid state %v for engine %s replica %s shallow copy", e.State, e.Name, destReplicaName)
+	}
+
+	if e.Frontend != types.FrontendSPDKVoid {
+		return fmt.Errorf("invalid frontend %v for engine %s replica %s shallow copy", e.Frontend, e.Name, destReplicaName)
```
derekbit, line 652: Should it be `FrontendEmpty`?

pkg/spdk/engine.go
```diff
@@ -498,6 +528,223 @@ func (e *Engine) validateAndUpdateReplicaMode(replicaName string, bdev *spdktype
...
+	for _, rpcSnapLvol := range rpcSrcReplica.Snapshots {
+		if rpcSnapLvol.Parent == "" {
+			ancestorSnapshotName = rpcSnapLvol.Name
+		}
+		if rpcSnapLvol.Children[srcReplicaName] {
+			latestSnapshotName = rpcSnapLvol.Name
+		}
+	}
+	if ancestorSnapshotName == "" || latestSnapshotName == "" {
+		return fmt.Errorf("cannot find the ancestor snapshot %s or latest snapshot %s from RW replica %s snapshot map during engine %s replica %s shallow copy", ancestorSnapshotName, latestSnapshotName, srcReplicaName, e.Name, destReplicaName)
```
derekbit, line 693: "failed ..."

pkg/spdk/server.go
```diff
@@ -346,10 +346,137 @@ func (s *Server) ReplicaSnapshotDelete(ctx context.Context, req *spdkrpc.Snapsho
...
+}
+
+func (s *Server) ReplicaSnapshotShallowCopy(ctx context.Context, req *spdkrpc.ReplicaSnapshotShallowCopyRequest) (ret *empty.Empty, err error) {
+	if req.Name == "" || req.SnapshotName == "" {
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "replica snapshot name is required")
+	}
+
+	s.RLock()
+	r := s.replicaMap[req.Name]
+	defer s.RUnlock()
```
derekbit, line 406: Why not

```suggestion
	s.RLock()
	defer s.RUnlock()

	r := s.replicaMap[req.Name]
```

pkg/spdk/server.go
```diff
@@ -466,6 +592,33 @@ func (s *Server) EngineWatch(req *empty.Empty, srv spdkrpc.SPDKService_EngineWat
...
+	}
+
+	s.Lock()
+	if err := e.ReplicaAddFinish(s.spdkClient, req.ReplicaName, req.ReplicaAddress, s.getLocalReplicaLvsNameMap(map[string]string{req.ReplicaName: ""})); err != nil {
+		s.Unlock()
+		return nil, err
+	}
+	s.Unlock()
+
+	return &empty.Empty{}, nil
```
derekbit, line 619: Can it be

```suggestion
	s.Lock()
    defer s.Unlock()

	if err := e.ReplicaAddFinish(s.spdkClient, req.ReplicaName, req.ReplicaAddress, s.getLocalReplicaLvsNameMap(map[string]string{req.ReplicaName: ""})); err != nil {
		return nil, err
	}

	return &empty.Empty{}, nil
```
?

pkg/spdk/engine.go
```diff
@@ -498,6 +528,223 @@ func (e *Engine) validateAndUpdateReplicaMode(replicaName string, bdev *spdktype
...
+
+	srcReplicaName, srcReplicaAddress := "", ""
+	for replicaName, replicaMode := range e.ReplicaModeMap {
+		if replicaMode != types.ModeRW {
+			continue
+		}
+		srcReplicaName = replicaName
+		srcReplicaAddress = e.ReplicaAddressMap[replicaName]
+		break
+	}
```
derekbit, line 663: Should we check if `srcReplicaName` and `srcReplicaAddress` are empty before shallow copy?

pkg/spdk/replica.go
```diff
@@ -691,11 +723,327 @@ func (r *Replica) removeLvolFromActiveChainWithoutLock(name string) int {
...
+	if err != nil {
+		return err
+	}
+	// TODO: After launching online rebuilding, the destination lvol name would be GetReplicaRebuildingLvolName(destReplicaName)
+	destRebuildingLvolName := destReplicaName
+
+	if destRebuildingLvolIP == r.IP {
+		destReplicaLvsName := localReplicaLvsNameMap[destReplicaName]
+		if destReplicaLvsName == "" {
+			return fmt.Errorf("cannot to find dest replica %s from the local replica map for replica %s rebuilding src start", destReplicaName, r.Name)
```
derekbit, line 765: "failed to ..."

pkg/spdk/engine.go
shuo-wu: Do we need to follow the same pattern as frontend `iscsi`? For example `iscsi://10.42.0.111:3260/iqn.2019-10.io.longhorn:vol/1`.

pkg/spdk/engine.go
shuo-wu: The PR for `types. FrontendSPDKVoid` is already merged hence I directly use this const. Do we need to rename it to `FrontendEmpty`

derekbit: I'm fine with `FrontendEmpty`. We can rename it.

pkg/spdk/engine.go
derekbit: Yes, maybe `nvmf://<ip>:<port>/<nqn>`?

pkg/spdk/engine.go
```diff
@@ -498,6 +528,223 @@ func (e *Engine) validateAndUpdateReplicaMode(replicaName string, bdev *spdktype
...
+func (e *Engine) ReplicaShallowCopy(destReplicaName, destReplicaAddress string) (err error) {
+	e.Lock()
+	defer e.Unlock()
+
+	// Syncing with the SPDK TGT server only when the engine is running.
+	if e.State != types.InstanceStateRunning {
+		return fmt.Errorf("invalid state %v for engine %s replica %s shallow copy", e.State, e.Name, destReplicaName)
+	}
+
+	if e.Frontend != types.FrontendSPDKVoid {
```
derekbit, line 651: Should it be `if e.Frontend != types.FrontendEmpty`?

pkg/spdk/engine.go
shuo-wu: Yes. It's better to have this check. Will add it

pkg/spdk/engine.go
shuo-wu: I am thinking if we need to differentiate the 2 cases:
1. Some errors fail the get/find. Then we use `failed to ...`
2. There is no such an entry in the map/list. Then we use `cannot find ...`

derekbit: Same here, because we check the words `cannot find ... ` in the longhorn-manager part for now and hould not always use `failed to ...`. 
We can improve the wording in the future.

shuo-wu: Sorry...? Which one should we use here?

derekbit: How about `cannot find ...`

shuo-wu: Sure. Then I will follow the convention I mentioned above.

derekbit (commented):
In general, LGTM except for the error handling.
As we discussed before, error handling will be addressed later.

pkg/client/client.go
```diff
@@ -229,21 +364,76 @@ func (c *SPDKClient) EngineWatch(ctx context.Context) (*api.EngineStream, error)
...
-	if err != nil {
-		return errors.Wrap(err, "failed to delete replica from engine")
+	return errors.Wrapf(err, "failed to add replica %s with address %s to engine %s", replicaName, replicaAddress, engineName)
+}
+
+func (c *SPDKClient) EngineReplicaDelete(engineName, replicaName, replicaAddress string) error {
+	if engineName == "" {
+		return fmt.Errorf("failed to delete replica from SPDK engine: missing required parameter engine name")
 	}
+	if replicaName == "" || replicaAddress == "" {
```
derekbit, line 423: Should be `if replicaName == "" && replicaAddress == "" {`
Either of the two parameters is enough.

pkg/spdk/types.go
```diff
@@ -1,16 +1,56 @@
...
 )
 
+func GetReplicaSnapshotLvolNamePrefix(replicaName string) string {
+	return fmt.Sprintf("%s-snap-", replicaName)
+}
+
 func GetReplicaSnapshotLvolName(replicaName, snapshotName string) string {
-	return fmt.Sprintf("%s-snaps-%s", replicaName, snapshotName)
+	return fmt.Sprintf("%s%s", GetReplicaSnapshotLvolNamePrefix(replicaName), snapshotName)
+}
```
derekbit, line 28: The max length of a lvol's name is 64 bytes.
If a name generated by `GetReplicaSnapshotLvolName` exceeds the number, the snapshot operation will fail.

derekbit: Can you update the max length (`SPDK_LVOL_NAME_MAX`) to a bigger number?
cc @DamiaSan

pkg/spdk/server.go
```diff
@@ -346,18 +346,145 @@ func (s *Server) ReplicaSnapshotDelete(ctx context.Context, req *spdkrpc.Snapsho
...
+}
+
+func (s *Server) ReplicaSnapshotShallowCopy(ctx context.Context, req *spdkrpc.ReplicaSnapshotShallowCopyRequest) (ret *empty.Empty, err error) {
+	if req.Name == "" || req.SnapshotName == "" {
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "replica snapshot name is required")
+	}
+
+	s.RLock()
+	r := s.replicaMap[req.Name]
+	defer s.RUnlock()
```
derekbit, line 406: Should be
```suggestion
	s.RUnlock()
```

derekbit:
The engine and automatic replica rebuilding integration is functioning properly. However, there is a potential deadlock issue within the current implementation.

The lock orders of the two functions are as follows:
```
verify(): s.Lock() -> e.Lock()
EngineReplicaAdd(): e.Lock() -> s.Lock()
```

This lock ordering occasionally leads to deadlocks.

To address this, we need to revise the locks in the SPDK engine.

pkg/spdk/types.go
derekbit: Handled in https://github.com/longhorn/spdk/pull/11

pkg/spdk/server.go
```diff
@@ -466,6 +592,32 @@ func (s *Server) EngineWatch(req *empty.Empty, srv spdkrpc.SPDKService_EngineWat
...
+		return nil, err
+	}
+
+	// Cannot add a lock for this call
+	if err := e.ReplicaShallowCopy(req.ReplicaName, req.ReplicaAddress); err != nil {
+		return nil, err
+	}
+
+	s.Lock()
+	defer s.Unlock()
```
derekbit, line 613: Can we remove `s.Lock()`?

pkg/spdk/server.go
```diff
@@ -346,18 +346,145 @@ func (s *Server) ReplicaSnapshotDelete(ctx context.Context, req *spdkrpc.Snapsho
...
+func (s *Server) ReplicaRebuildingDstSnapshotCreate(ctx context.Context, req *spdkrpc.SnapshotRequest) (ret *empty.Empty, err error) {
+	if req.Name == "" || req.SnapshotName == "" {
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "replica name and snapshot name are required")
+	}
+
+	s.Lock()
+	defer s.Unlock()
+
+	r := s.replicaMap[req.Name]
+
```
derekbit, line 469: Can we use a map lock for reading/updating replicaMap and engineMap?
```
	s.mapLock.RLock()
	r := s.replicaMap[req.Name]
	s.mapLock.RUnlock()
```

Same question for other server's methods.

pkg/spdk/types.go
shuo-wu: Not sure if upstream would accept this change... Do we need to truncate `replicaName` if it's too long?

shuo-wu:
@derekbit I will make `s.spdkClient` thread safe then `s.Lock` is responsible for protecting the maps only. WDYT?

derekbit:
> @derekbit I will make `s.spdkClient` thread safe then `s.Lock` is responsible for protecting the maps only. WDYT?

Sounds good to me.

pkg/spdk/types.go
derekbit: > Do we need to truncate replicaName if it's too long?

I'm fine with the idea.
However, 64 bytes is really short. The NAME_MAX on Linux is 255 ([ref](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/limits.h#L12)). Increasing the number is reasonable to me.

@DamiaSan Can you help check the change with SPDK maintainers? Thank you.

innobead: Let's do a workaround here first by truncating if no extra concerns, but at the same, see if 255 name length is feasible for upstream. @shuo-wu @derekbit ?

pkg/client/client.go
derekbit: Hello @shuo-wu 
Forgot update it.

derekbit (approved):
LGTM.
As discussed before, the lock issue will be addressed in a separate PR.

pkg/spdk/replica.go
```diff
@@ -649,53 +689,376 @@ func (r *Replica) SnapshotDelete(spdkClient *spdkclient.Client, snapshotName str
...
+			return err
+		}
+		if len(nvmeBdevNameList) != 1 {
+			return fmt.Errorf("got zero or multiple results when attaching rebuilding dst lvol %s with address %s as a NVMe bdev: %+v", dstRebuildingLvolName, dstRebuildingLvolAddress, nvmeBdevNameList)
+		}
+		r.rebuildingDstBdevName = nvmeBdevNameList[0]
+	}
+
+	r.rebuildingDstReplicaName = dstReplicaName
+	r.rebuildingDstBdevType = spdktypes.BdevTypeNvme
```
innobead, line 787: When will `rebuildingDstBdevType` be `spdktypes.BdevTypeLvol`? I thought it should be rebuilding happening on the same node, right? Just noticed you handle this case in RebuildingSrcFinish.

pkg/spdk/replica.go
```diff
@@ -649,53 +689,376 @@ func (r *Replica) SnapshotDelete(spdkClient *spdkclient.Client, snapshotName str
...
+		}
+		r.IsExposed = false
+	}
+	r.rebuildingPort = 0
+	r.rebuildingLvol = nil
+
+	// TODO: For online rebuilding, connect the snapshot tree/chain with the head lvol then mark the head lvol as mode RW
+
+	// Then re-construct the replica ActiveChain (and SnapshotMap)
+	// TODO: Can we somehow rely on the periodic sync to re-construct the chain?
```
innobead, line 979: If we accept eventual consistency, it should be feasible and it will be good to have one place to do the chain construction. 

However, the primary concern would be consistency like rebuilding has been finished, but the chain is not updated yet?

pkg/spdk/server.go
```diff
@@ -466,6 +587,32 @@ func (s *Server) EngineWatch(req *empty.Empty, srv spdkrpc.SPDKService_EngineWat
 	return nil
 }
 
+func (s *Server) EngineReplicaAdd(ctx context.Context, req *spdkrpc.EngineReplicaAddRequest) (ret *empty.Empty, err error) {
```
innobead, line 590: So, this operation is blocking. Will timeout be a concern?

innobead (dismissed):
In general, LGTM.

Some comments were left. Need to wait for @shuo-wu testing before merging this PR.

innobead:
@shuo-wu Please create the following issues for the following enhancements also including TODO items.

pkg/spdk/engine.go
```diff
@@ -498,6 +528,227 @@ func (e *Engine) validateAndUpdateReplicaMode(replicaName string, bdev *spdktype
...
+		return fmt.Errorf("invalid frontend %v for engine %s replica %s add start", e.Frontend, e.Name, replicaName)
+	}
+
+	if _, exists := e.ReplicaAddressMap[replicaName]; exists {
+		return fmt.Errorf("replica %s already exists", replicaName)
+	}
+
+	defer func() {
+		if err != nil && e.State != types.InstanceStateError {
+			e.State = types.InstanceStateError
```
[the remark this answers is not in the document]
shuo-wu: Yes. Will raise a separate PR for error message recording

pkg/spdk/replica.go
```diff
@@ -166,7 +176,14 @@ func (r *Replica) construct(bdevLvolMap map[string]*spdktypes.BdevInfo) (err err
...
 	}()
 
-	if r.State != types.InstanceStatePending {
+	switch r.State {
+	case types.InstanceStatePending:
+		break
+	case types.InstanceStateRunning:
+		if r.isRebuilding {
+			break
+		}
```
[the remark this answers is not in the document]
shuo-wu: No, I missed `fallthrough` here. If it's not a rebuilding replica, this should not continue.

pkg/spdk/replica.go
shuo-wu: Yes. When src replica and dst replica are on the same node, we cannot do nvme_controller_attach. Instead, we should directly use the local lvol bdev.

pkg/spdk/replica.go
shuo-wu: Good point! Then it's better to construct the chain separately, before the rebuilding related fields are cleaned up.
I will remove this comment

pkg/spdk/server.go
shuo-wu: Yes. We should use longer timeouts for this API and `ReplicaSnapshotShallowCopy`

innobead: Should think about an async way later after 1.5.0 to make us support longer rebuild time.

pkg/spdk/engine.go
```diff
@@ -465,6 +412,87 @@ func (e *Engine) ValidateAndUpdate(
...
+	if e.Frontend != types.FrontendEmpty && e.Frontend != types.FrontendSPDKTCPNvmf && e.Frontend != types.FrontendSPDKTCPBlockdev {
+		return fmt.Errorf("unknown frontend type %s", e.Frontend)
+	}
+
+	nqn := helpertypes.GetNQN(e.Name)
+	subsystem := subsystemMap[nqn]
+
+	if e.Frontend == types.FrontendEmpty {
+		if subsystem != nil {
+			return fmt.Errorf("found nvmf subsystem %s for engine %s with void frontend", nqn)
```
derekbit, line 425:
```suggestion
			return fmt.Errorf("found nvmf subsystem %s for engine %s with void frontend", nqn, e.Name)
```

pkg/spdk/engine.go
```diff
@@ -465,6 +412,87 @@ func (e *Engine) ValidateAndUpdate(
...
+		}
+		if e.Endpoint != "" {
+			return fmt.Errorf("found non-empty endpoint %s for engine %s with void frontend", e.Endpoint)
```
derekbit, line 428:
```suggestion
			return fmt.Errorf("found non-empty endpoint %s for engine %s with void frontend", e.Endpoint, e.Name)
```

pkg/spdk/engine.go
```diff
@@ -465,6 +412,87 @@ func (e *Engine) ValidateAndUpdate(
...
+		}
+		if e.Port != 0 {
+			return fmt.Errorf("found non-zero port %s for engine %s with void frontend", e.Port)
```
derekbit, line 431:
```suggestion
			return fmt.Errorf("found non-zero port %d for engine %s with void frontend", e.Port, e.Name)
```

pkg/spdk/engine.go
```diff
@@ -504,6 +532,230 @@ func (e *Engine) validateAndUpdateReplicaMode(replicaName string, bdev *spdktype
...
+		return err
+	}
+	if err = srcReplicaServiceCli.ReplicaRebuildingSrcStart(srcReplicaName, dstReplicaName, dstReplicaAddress); err != nil {
+		return err
+	}
+
+	// Reverse the src replica snapshot tree with a DFS way and do shallow copy one by one
+	stack := []string{ancestorSnapshotName}
+	for currentSnapshotName := stack[len(stack)-1]; len(stack) > 0; {
+		currentSnapshotName = stack[len(stack)-1]
```
derekbit, line 725:
```
currentSnapshotName = stack[len(stack)-1]
```
Is it redundant?

innobead:
> verify(): s.Lock() -> e.Lock()
> EngineReplicaAdd(): e.Lock() -> s.Lock()

For this kind of concurrency, it's better to use the same lock to do the concurrency, for example:

- verify: s.Lock() -> e.Lock(), which implicitly happens in the engine
- EngineReplicaAdd -> s.Lock() -> e.Lock()

Also, in general, when dealing with concurrency, in some cases, can also think about the message-passing pattern which could prevent the potential race condition. ref: https://www.baeldung.com/concurrency-principles-patterns

pkg/client/client.go
```diff
@@ -229,21 +366,76 @@ func (c *SPDKClient) EngineWatch(ctx context.Context) (*api.EngineStream, error)
...
+	defer cancel()
+
+	_, err := client.EngineSnapshotDelete(ctx, &spdkrpc.SnapshotRequest{
+		Name:         name,
+		SnapshotName: snapshotName,
+	})
+	return errors.Wrapf(err, "failed to delete SPDK engine %s snapshot %s", name, snapshotName)
+}
+
+func (c *SPDKClient) EngineReplicaAdd(engineName, replicaName, replicaAddress string) error {
```
innobead, line 401: It's better to make time-consuming operations async. This can be improved later. Basically, have a call operation, then stats operation to understand the status.

innobead (commented):
In general, LGTM. 

Let's do more testing and the remaing part can be improved in later versions. Also, need to resolve Derek's reviews.

derekbit:
SPDK volumes attachment/detachment cannot function properly because of similar errors.
```
[longhorn-instance-manager] time="2023-06-12T16:27:03Z" level=error msg="Failed to get disk info" diskName=test-block-1 error="rpc error: code = Internal desc = failed to get AIO bdev with name test-block-1: error sending message, id 70446, method bdev_get_bdevs, params {test-block-1 0}: invalid character 'e' looking for beginning of value"
```

I guess it is related to SPDK JSON client that is currently not thread-safe.
The issue is gone after applying a spdk client lock.
Shall we add a spdk client lock for `rc2` before introducing a thread-safe client?

cc @innobead @shuo-wu

innobead:
Agreed. Locking is also a way to achieve thread safety, even though need to handle it carefully.

pkg/client/client.go
shuo-wu: The legacy rebuilding API is a sync call as well. It's better to make the change simultaneously.

innobead: I see. Let's improve it later.

innobead (approved):

derekbit (approved):
LGTM.

PR #50, opened 2023-06-10
Ignore error due to controller's connecting or resetting state

derekbit:
The error is recoverable and no need to mark the engine as error.

waiting for https://github.com/longhorn/go-spdk-helper/pull/37

Longhorn/longhorn#5744

### DIFF
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -353,7 +353,13 @@ func (e *Engine) ValidateAndUpdate(
 		if err != nil {
 			return err
 		}
+
 		if err := initiator.LoadNVMeDeviceInfo(); err != nil {
+			if strings.Contains(err.Error(), "connecting state") ||
+				strings.Contains(err.Error(), "resetting state") {
+				e.log.WithError(err).Warnf("Ignored to validate and update engine %v, because the device is still in a transient state", e.Name)
+				return nil
+			}
 			return err
 		}
 
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.17
 require (
 	github.com/RoaringBitmap/roaring v1.2.3
 	github.com/golang/protobuf v1.5.2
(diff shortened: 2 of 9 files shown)


pkg/spdk/engine.go
```diff
@@ -353,7 +353,13 @@ func (e *Engine) ValidateAndUpdate(
 		if err != nil {
 			return err
 		}
+
 		if err := initiator.LoadNVMeDeviceInfo(); err != nil {
+			if strings.Contains(err.Error(), "connecting state") ||
```
innobead, line 358: nit: determining different errors by parsing string would be potentially error-prone. I would suggest having gRPC error like each error still implements the error interface but also has a specific interface to provide error code capability. 

This can be a way to improve our error handling, especially need to differentiate different error types.

pkg/spdk/engine.go
```diff
@@ -353,7 +353,13 @@ func (e *Engine) ValidateAndUpdate(
...
+				strings.Contains(err.Error(), "resetting state") {
+				e.log.WithError(err).Warnf("Failed to validate and update engine %s", e.Name)
```
innobead, line 360: `Ignored to valide and update engine..., because the device is still in a transient state.`

innobead (changes requested):

pkg/spdk/engine.go
derekbit: Personally, I like to use error code. However, it seems there is no good error code lib we can use directly.
We may can implement an error interface lib for our applications.

pkg/spdk/engine.go
derekbit: Updated, thanks.

innobead (approved):
LGTM.

pkg/spdk/engine.go
innobead: Sounds good. Let's create an issue for this improvement later.

derekbit: Tracked in https://github.com/longhorn/longhorn/issues/6093

PR #51, opened 2023-06-13
 Protect SPDK client by a mutex lock

derekbit:
Waiting for https://github.com/longhorn/longhorn-spdk-engine/pulls

Longhorn/longhorn#6067
Longhorn/longhorn#6071

### DIFF
--- a/pkg/spdk/client.go
+++ b/pkg/spdk/client.go
@@ -1,0 +1,161 @@
+package spdk
+
+import (
+	"sync"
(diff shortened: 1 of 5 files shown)


innobead (approved):
LGTM.

This is a temporary solution. Eventually, we should make the SPDK client thread safe which means requing to figure out a way to deal with the communication (response part) with the SPDK RPC server.

PR #52, opened 2023-06-14
Set transport_ack_timeout for recognizing TCP disconnection

derekbit:
Waiting for https://github.com/longhorn/go-spdk-helper/pull/38
longhorn/longhorn#6110

### DIFF
--- a/pkg/spdk/server.go
+++ b/pkg/spdk/server.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	grpccodes "google.golang.org/grpc/codes"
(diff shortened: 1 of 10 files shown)


pkg/spdk/server.go
```diff
@@ -46,6 +48,11 @@ func NewServer(ctx context.Context, portStart, portEnd int32) (*Server, error) {
 		return nil, err
 	}
 
+	if _, err = cli.BdevNvmeSetOptions(helpertypes.DefaultCtrlrLossTimeoutSec, helpertypes.DefaultReconnectDelaySec,
```
innobead, line 51: nit: For better formatting, you can write this way instead.

```go
err = cli.BdevNvmeSetOptions(
    helpertypes.DefaultCtrlrLossTimeoutSec, 
    helpertypes.DefaultReconnectDelaySec, 
    helpertypes.DefaultFastIOFailTimeoutSec, 
    helpertypes.DefaultTransportAckTimeout,
)
if err != nil {
    return nil, errors.Wrap(err, "failed to set nvme options")
}
```

innobead (commented):
LGTM, just pending for spdk helper PR merge.

pkg/spdk/server.go
derekbit: Updated, thank you.

shuo-wu (approved):

PR #53, opened 2023-06-14
disk: add diskUUID in DiskCreate

derekbit:
If a disk is added before, the diskUUID is be recorded in node.status.diskStatus. The recorded UUID can be used for validating if the disk is an expected one.

Longhorn/longhorn#6125


pkg/spdk/disk.go
```diff
@@ -266,15 +268,23 @@ func addBlockDevice(spdkClient *SPDKClient, diskName, diskPath string, blockSize
...
-		renamed, err := spdkClient.BdevLvolRenameLvstore(lvstore.Name, lvstoreName)
-		if err != nil {
-			return "", errors.Wrapf(err, "failed to rename lvstore from %v to %v", lvstore.Name, lvstoreName)
-		}
-		if !renamed {
-			return "", fmt.Errorf("failed to rename lvstore from %v to %v", lvstore.Name, lvstoreName)
+		// Rename the existing lvstore to the name of the aio bdev
+		if diskUUID == lvstore.UUID {
+			log.Infof("Renaming the existing lvstore %v to %v", lvstore.Name, lvstoreName)
+			renamed, err := spdkClient.BdevLvolRenameLvstore(lvstore.Name, lvstoreName)
```
shuo-wu, line 274:
```
if lvstore.Name != lvstoreName {
    renamed, err := spdkClient.BdevLvolRenameLvstore(lvstore.Name, lvstoreName)
    ......
}
```

pkg/spdk/disk.go
```diff
@@ -266,15 +268,23 @@ func addBlockDevice(spdkClient *SPDKClient, diskName, diskPath string, blockSize
...
 		}
-		return lvstore.UUID, nil
+	}
+
+	// The lvstore should be created before, but it cannot be found now.
+	if diskUUID != "" {
+		return "", grpcstatus.Error(grpccodes.NotFound, fmt.Sprintf("cannot find lvstore with UUID %v", diskUUID))
 	}
 
 	return spdkClient.BdevLvolCreateLvstore(lvstoreName, diskName, defaultClusterSize)
```
shuo-wu, line 290: NIT: To make this function simpler, you can handle this (`diskUUID == ""`) first

pkg/spdk/disk.go
```diff
@@ -283,7 +293,7 @@ func addBlockDevice(spdkClient *SPDKClient, diskName, diskPath string, blockSize
 func lvstoreToDisk(spdkClient *SPDKClient, diskPath, lvstoreName, lvstoreUUID string) (*spdkrpc.Disk, error) {
 	lvstores, err := spdkClient.BdevLvolGetLvstore(lvstoreName, lvstoreUUID)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get lvstore with name %v and UUID %v", lvstoreName, lvstoreUUID)
+		return nil, errors.Wrapf(err, "cannot find lvstore with name %v and UUID %v", lvstoreName, lvstoreUUID)
```
shuo-wu, line 296: No need to update this log. This is an API invocation failure rather than not finding an entry in a valid list/map

pkg/spdk/disk.go
derekbit: Sorry. The comment is wrong.
The idea here is only renaming the lvstore which has an expected uuid.

pkg/spdk/disk.go
derekbit: Reverted. Thank you.

pkg/spdk/disk.go
derekbit: Updated. Thank you.

pkg/spdk/disk.go
derekbit: Updated the comment, thank you.

pkg/spdk/disk.go
```diff
@@ -266,18 +268,26 @@ func addBlockDevice(spdkClient *SPDKClient, diskName, diskPath string, blockSize
...
+			}
+			return lvstore.UUID, nil
 		}
-		return lvstore.UUID, nil
 	}
 
-	return spdkClient.BdevLvolCreateLvstore(lvstoreName, diskName, defaultClusterSize)
+	if diskUUID == "" {
+		return spdkClient.BdevLvolCreateLvstore(lvstoreName, diskName, defaultClusterSize)
+	}
```
shuo-wu, line 287: I mean:
```
	if diskUUID == "" {
		return spdkClient.BdevLvolCreateLvstore(lvstoreName, diskName, defaultClusterSize)
	}
	
	for _, lvstore := range lvstores {
		if lvstore.BaseBdev != bdevName {
			continue
		}
		if lvstore.UUID != diskUUID {
			continue
		}
		if lvstore.Name == lvstoreName {
			return lvstore.UUID, nil
		}
		log.Infof("Renaming the existing lvstore %v to %v", lvstore.Name, lvstoreName)
		renamed, err := spdkClient.BdevLvolRenameLvstore(lvstore.Name, lvstoreName)
		if err != nil {
			return "", errors.Wrapf(err, "failed to rename lvstore from %v to %v", lvstore.Name, lvstoreName)
		}
	}

	return "", grpcstatus.Error(grpccodes.NotFound, fmt.Sprintf("cannot find lvstore with UUID %v", diskUUID))
```

derekbit: Updated, thanks.

shuo-wu (approved):

proto/spdkrpc/spdk.proto
```diff
@@ -203,10 +203,11 @@ message Disk {
...
 }
 
-message DiskCreateRequest{
+message DiskCreateRequest {
 	string disk_name = 1;
-    string disk_path= 2;
-    int64 block_size = 3;
+    string disk_uuid = 2;
+    string disk_path= 3;
+    int64 block_size = 4;
```
innobead, line 210: Fix the formatting.

pkg/spdk/disk.go
```diff
@@ -261,11 +268,16 @@ func addBlockDevice(spdkClient *SPDKClient, diskName, diskPath string, blockSize
 			continue
 		}
 
+		if diskUUID != lvstore.UUID {
+			continue
```
innobead, line 272: so the UUID is actually lvstore UUID.

innobead (changes requested):
Just one format comment needed to fix.

pkg/spdk/disk.go
derekbit: Yes, you're right

proto/spdkrpc/spdk.proto
derekbit: Fixed. Thank you.

innobead (approved):

PR #55, opened 2023-06-19
Improve resilience to spdk_tgt crash

derekbit:
- Mark non-stopped and non-error replicas and engines as error when spdk_tgt is down 
    When the spdk_tgt is somehow down, the replicas and engines are out of sync, need to mark non-stopped and non-error replicas and engines as error. Then, longhorn-manager can be aware of the failed replicas and engines and clean them up.

- Restart spdk_tgt and reconnect to it if it is somehow down

longhorn/longhorn#6071
longhorn/longhorn#6155

### DIFF
--- a/pkg/spdk/disk.go
+++ b/pkg/spdk/disk.go
@@ -114,12 +114,12 @@ func svcDiskGet(spdkClient *SPDKClient, diskName string) (ret *spdkrpc.Disk, err
 		"diskName": diskName,
 	})
 
-	log.Info("Getting disk info")
+	log.Trace("Getting disk info")
 	defer func() {
 		if err != nil {
 			log.WithError(err).Error("Failed to get disk info")
 		} else {
-			log.Info("Got disk info")
+			log.Trace("Got disk info")
 		}
 	}()
 
--- a/pkg/spdk/server.go
+++ b/pkg/spdk/server.go
@@ -114,7 +114,7 @@ func (s *Server) monitoring() {
 	}
 }
 
-func (s *Server) verify() error {
+func (s *Server) verify() (err error) {
 	replicaMap := map[string]*Replica{}
 	engineMap := map[string]*Engine{}
 	s.RLock()
@@ -125,6 +125,23 @@ func (s *Server) verify() error {
 		engineMap[k] = v
 	}
 	s.RUnlock()
+
+	defer func() {
+		if err == nil {
+			return
+		}
+		if jsonrpc.IsJSONRPCRespErrorBrokenPipe(err) {
+			logrus.WithError(err).Errorf("SPDK Server: marking all non-stopped and non-error replicas and engines as error")
+			s.RLock()
+			for _, r := range s.replicaMap {
+				r.SetErrorState()
+			}
+			for _, e := range s.engineMap {
+				e.SetErrorState()
+			}
+			s.RUnlock()
+		}
+	}()
 
 	lvsList, err := s.spdkClient.BdevLvolGetLvstore("", "")
 	if err != nil {
--- a/pkg/spdk/server.go
+++ b/pkg/spdk/server.go
@@ -13,6 +13,7 @@ import (
 	grpcstatus "google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/emptypb"
 
+	"github.com/longhorn/go-spdk-helper/pkg/jsonrpc"
 	spdktypes "github.com/longhorn/go-spdk-helper/pkg/spdk/types"
 	helpertypes "github.com/longhorn/go-spdk-helper/pkg/types"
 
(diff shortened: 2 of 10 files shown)


innobead (approved):
LGTM. Pending to merge until the go-spdk-helper PR merged.

pkg/spdk/server.go
```diff
@@ -126,6 +126,23 @@ func (s *Server) verify() error {
 	}
 	s.RUnlock()
 
+	defer func() {
+		if err == nil {
+			return
+		}
+		if jsonrpc.IsJSONRPCRespErrorBrokenPipe(err) {
+			logrus.WithError(err).Errorf("SPDK Server: marking all non-stopped and non-error replicas and engines as error")
+			s.RLock()
```
innobead, line 169: `.Error(..` instead.

innobead (changes requested):

pkg/spdk/server.go
```diff
@@ -104,8 +105,41 @@ func (s *Server) monitoring() {
...
+					logrus.WithError(err).Errorf("SPDK Server: failed to reconnect to spdk_tgt")
+					break
+				}
+
+				err = s.spdkClient.Close()
+				if err != nil {
+					logrus.WithError(err).Warnf("SPDK Server: failed to close spdk_tgt connection")
+				}
+				s.spdkClient = cli
+			} else {
```
innobead, line 138: lock to protect `s.spdkClient`?

pkg/spdk/server.go
```diff
@@ -104,8 +105,41 @@ func (s *Server) monitoring() {
...
+				err = util.StartSPDKTgtDaemon()
+				if err != nil {
+					logrus.WithError(err).Errorf("SPDK Server: failed to start spdk_tgt daemon")
+				}
 			}
```
innobead, line 143: How about moving this code to a private method instead? like `tryEnsureSPDKTgtHealthy`

innobead (changes requested):
Missed one commit. There are a few comments to resolve.

pkg/spdk/server.go
derekbit: Is a lock necessary?
Without a lock, some operations might use broken spdkClient, but they will use the new one next time.

innobead: Yeah, since the client is already broken. It should be fine for now, but ideally, for a shared resource, it's going to change during the potential usage, it's better to protect it to prevent it has been changed in between. Hower, it's fine here.

pkg/spdk/server.go
derekbit: Addressed, thank you.

pkg/spdk/server.go
derekbit: Addressed, thank you.

pkg/spdk/server.go
derekbit: Refactored it and added a lock protection

pkg/spdk/server.go
```diff
@@ -114,6 +125,21 @@ func (s *Server) monitoring() {
...
+		return errors.Wrap(err, "failed to check spdk_tgt is running")
+	}
+
+	if running {
+		logrus.Info("SPDK Server: reconnecting to spdk_tgt")
+		return s.spdkClient.Reconnect()
+	}
+
+	logrus.Info("SPDK Server: restarting spdk_tgt")
+	return util.StartSPDKTgtDaemon()
```
shuo-wu, line 140: Does this mean that the instance manager (daemon) is responsible for the first time spdk_tgt start but the spdk engine server will take care of the following spdk_tgt restart? This looks a little bit weird...

shuo-wu (changes requested):

pkg/spdk/server.go
derekbit: Yes.
But longhorn-engine handles/restarts `tgtd` using the same weird pattern... Let's improve them in the future.
https://github.com/longhorn/longhorn-engine/blob/master/vendor/github.com/longhorn/go-iscsi-helper/iscsidev/iscsi.go#L88

pkg/spdk/server.go
```diff
@@ -126,6 +126,23 @@ func (s *Server) verify() error {
...
+		}
+		if jsonrpc.IsJSONRPCRespErrorBrokenPipe(err) {
+			logrus.WithError(err).Errorf("SPDK Server: marking all non-stopped and non-error replicas and engines as error")
+			s.RLock()
+			for _, r := range s.replicaMap {
+				r.SetErrorState()
+			}
+			for _, e := range s.engineMap {
+				e.SetErrorState()
+			}
```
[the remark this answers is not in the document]

pkg/spdk/server.go
innobead: Let's make it to the same place in a later version.

pkg/spdk/disk.go
```diff
@@ -114,12 +114,12 @@ func svcDiskGet(spdkClient *SPDKClient, diskName string) (ret *spdkrpc.Disk, err
 		"diskName": diskName,
 	})
 
-	log.Info("Getting disk info")
+	log.Trace("Getting disk info")
 	defer func() {
```
innobead, line 118: Good change.

pkg/util/util.go
```diff
@@ -43,3 +47,38 @@ func RemovePrefix(path, prefix string) string {
 func UUID() string {
 	return uuid.New().String()
 }
+
+func IsSPDKTargetProcessRunning() (bool, error) {
+	cmd := exec.Command("pgrep", "-f", "spdk_tgt")
```
innobead, line 52: nit: it would be good to not rely on the system cli but go native way.

innobead (approved):
LGTM.

PR #66, opened 2023-11-03
Add a linear dm device on the top of a nvme device

derekbit:
longhorn/longhorn#6001

deps on https://github.com/longhorn/go-spdk-helper/pull/50


derekbit:
Close the PR and handled it in v2 engine upgrade together.

PR #68, opened 2023-12-05
Add shutdown channel

derekbit:
After receiving a shutdown singal, the spdk gRPC server should stop.

Longhorn/longhorn#7263

### DIFF
--- a/pkg/spdk/server.go
+++ b/pkg/spdk/server.go
@@ -31,7 +31,8 @@ const (
 type Server struct {
 	sync.RWMutex
 
-	ctx context.Context
+	ctx        context.Context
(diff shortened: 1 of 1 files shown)


innobead (approved):

shuo-wu:
I am thinking why not use a child cancel context for the server instead

derekbit:
> I am thinking why not use a child cancel context for the server instead

It is also a solution. Other gRPC servers use shutdown channel, so I just use the same interface here.

derekbit:
> OK... I prefer to use a single context, which makes the implementation simpler

Sure. I'm thinking how to make all servers' interfaces simpler.

PR #70, opened 2023-12-07
Revert "Add shutdown channel"

derekbit:
This reverts commit 7bccdcac2687665c8cd38467c34cee62e7c848a3.

As the suggestion of @shuo-wul, use context method. Other gRPC services will use context rather than shutdown channel as well.

Longhorn/longhorn#7263


innobead (approved):
The context for this revert is that we would like to unify the callback by context later.

innobead (approved):

PR #72, opened 2023-12-07
Add StopSPDKTgtDaemon method

derekbit:
longhorn/longhorn#7263


derekbit:
Ready for review.

cc @innobead @shuo-wu @c3y1huang

derekbit:
Not sure why the uuid-devel is not found on aarch64 repo. @shuo-wu Do you have any idea?

pkg/util/util.go
```diff
@@ -86,6 +91,41 @@ func StartSPDKTgtDaemon() error {
...
+	var errs error
+	for _, process := range processes {
+		if err := process.Signal(syscall.SIGTERM); err != nil {
+			multierr.Append(errs, errors.Wrapf(err, "failed to send SIGTERM to spdk_tgt %v", process.Pid))
+		} else {
+			done := make(chan error, 1)
+			go func() {
+				_, err := process.Wait()
+				done <- err
+				close(done)
```
innobead, line 109: nit: Saw the discussion before about if you need to close the channel. In this case, it's not needed, because it's scoped, and you already sent the err to the channel.

innobead (approved):
LGTM

PR #75, opened 2023-12-13
Blindly stop raid bdev exposure before exposing it

derekbit:
In some failed cases, SPDK engine raises an error `nvmf ... subsystem already exists` while attempting to create a new subsystem for a RAID bdev. This is due to orphaned resources remaining undeleted after encountering an error.

To prevent this error message, we can blindly stop raid bdev exposure for cleaning up the orphaned resources before exposing it.

longhorn/longhorn#7324

### DIFF
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	"github.com/longhorn/go-spdk-helper/pkg/jsonrpc"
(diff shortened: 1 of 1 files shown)


pkg/spdk/engine.go
```diff
@@ -173,6 +174,12 @@ func (e *Engine) handleFrontend(spdkClient *spdkclient.Client, portCount int32,
 	}
 
 	nqn := helpertypes.GetNQN(e.Name)
+
+	e.log.Infof("Blindly stopping expose bdev for engine")
+	if err := spdkClient.StopExposeBdev(nqn); err != nil {
```
innobead, line 179: ref: if not found, there will be no error returned.

[/pkg/spdk/client/advanced.go#L92-L94](https://github.com/longhorn/go-spdk-helper/blob/3b72caa5a1eac0e26f92db346009f5a9c5f7fa0e/pkg/spdk/client/advanced.go#L92-L94)
```go
	if subsystem == nil {
		return nil
	}
```

innobead (approved):
LGTM

PR #88, opened 2024-01-06
Make the format of log messages consistent

derekbit:

Issue #

Make the format of log messages consistent

### DIFF
--- a/pkg/spdk/backup.go
+++ b/pkg/spdk/backup.go
@@ -10,6 +10,8 @@ import (
 	"github.com/0xPolygon/polygon-edge/consensus/polybft/bitmap"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	grpccodes "google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
 
 	"github.com/longhorn/backupstore"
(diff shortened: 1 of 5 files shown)


innobead (changes requested):
Suggest using gRPC errors exclusively within the gRPC server layer. For internal operations, let's continue using the usual error mechanism. This way will ensure clarity regarding when and where to use gRPC errors. WDYT?

derekbit:
> Suggest using gRPC errors exclusively within the gRPC server layer. For internal operations, let's continue using the usual error mechanism. This way will ensure clarity regarding when and where to use gRPC errors. WDYT?

I thought the method before. There are many checks in the engine.go, replica.go and so on. The checks are related to different gRPC error codes, so using gRPC errors in server layer is not enough.

innobead:
I see. Let's do this change first, but we should adopt the suggestion later, or the error handling code would become hard to maintain eventually.

innobead:
Have a todo.

innobead (approved):

derekbit:
Make it as a draft. Handle it in v1.7.0.

innobead:
👍

derekbit:
Handle this later.

PR #98, opened 2024-01-12
SPDK engine is responsible for collecting the snapshot info and actual size

shuo-wu:
This PR fixes part of this issue: https://github.com/longhorn/longhorn/issues/7641


pkg/api/types.go
```diff
@@ -46,9 +49,12 @@ func ProtoLvolToLvol(l *spdkrpc.Lvol) *Lvol {
...
 
 func LvolToProtoLvol(l *Lvol) *spdkrpc.Lvol {
+	if l == nil {
+		return nil
+	}
 	return &spdkrpc.Lvol{
-		Name:         l.Name,
-		Uuid:         l.UUID,
+		Name: l.Name,
+		// Uuid:         l.UUID,
```
derekbit, line 57: Can we add `UserCreated` flag? Currently, the flag is always true before finishing  https://github.com/longhorn/longhorn/issues/7578 .

shuo-wu: Do we need to handle this field in the current PR? I prefer to tackle it later.

derekbit: Sure!

derekbit (approved):
LGTM

derekbit:
@shuo-wu Will you update instance-manager after getting merged?

innobead:
LGTM

pkg/api/types.go
innobead: @DamiaSan is working on it on SPDK part, so he should be able to help with longhorn integration here.

PR #101, opened 2024-01-17
Updates for dm deferred deletion option and the ctrl-loss-tmo of "nvme connect"

derekbit:

Issue https://github.com/longhorn/longhorn/issues/7697

- Allow derferred dm device deletion accordingly
    When deleting a replica, no need to delete a dm device by a blocking method.

- Pass DefaultKeepAliveTimeoutMs to BdevNvmeSetOptions

Deps on https://github.com/longhorn/go-spdk-helper/pull/70

### DIFF
--- a/pkg/spdk/backup.go
+++ b/pkg/spdk/backup.go
@@ -225,7 +225,7 @@ func (b *Backup) CloseSnapshot(snapshotName, volumeName string) error {
 	}
 
 	b.log.Info("Stopping NVMe initiator")
-	if _, err := b.initiator.Stop(false, true); err != nil {
+	if _, err := b.initiator.Stop(false, false, true); err != nil {
 		return errors.Wrapf(err, "failed to stop NVMe initiator")
 	}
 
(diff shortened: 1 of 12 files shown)


pkg/spdk/backup.go
```diff
@@ -225,7 +225,7 @@ func (b *Backup) CloseSnapshot(snapshotName, volumeName string) error {
 	}
 
 	b.log.Info("Stopping NVMe initiator")
-	if _, err := b.initiator.Stop(false, true); err != nil {
+	if _, err := b.initiator.Stop(false, false, true); err != nil {
```
innobead, line 228: `false, true, true`? the second parameter should be `isDeferred`?

innobead (changes requested):
Just one question, the same as one left in go-spdk-helper.

After clarification and update the dep, we can merge this.

pkg/spdk/backup.go
derekbit: Either `false, true, true` or `false, false, true` is fine, because backup doesn't use dm linear device.

innobead (approved):

PR #103, opened 2024-01-18
spdk: Blindly finish rebuilding for src replica after rebuilding failure

shuo-wu:
After rebuilding failed, the previous rebuilding info in the source replica won't be cleaned up. Then the following rebuilding cannot use that source replica anymore.

Issue https://github.com/longhorn/longhorn/issues/7723

### DIFF
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -938,6 +938,10 @@ func (e *Engine) ReplicaShallowCopy(dstReplicaName, dstReplicaAddress string) (e
 	defer func() {
 		// Blindly mark the rebuilding replica as mode ERR now.
 		if err != nil {
+			// Blindly send rebuilding finish for src replica.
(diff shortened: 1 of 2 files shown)


pkg/spdk/engine.go
```diff
@@ -938,6 +938,10 @@ func (e *Engine) ReplicaShallowCopy(dstReplicaName, dstReplicaAddress string) (e
 	defer func() {
 		// Blindly mark the rebuilding replica as mode ERR now.
 		if err != nil {
+			// Blindly send rebuilding finish for src replica.
+			if srcReplicaErr := srcReplicaServiceCli.ReplicaRebuildingSrcFinish(srcReplicaName, dstReplicaName); srcReplicaErr != nil {
```
innobead, line 942: Do you want to wrap `srcReplicaErr` as well? or it's meaningless here?

shuo-wu: Typo below. Will just print out it.

innobead (approved):

PR #105, opened 2024-01-29
Abstract disk operations for different disk drivers

derekbit:

Issue longhorn/longhorn#7672

Deps on https://github.com/longhorn/go-spdk-helper/pull/72


pkg/spdk/disk/types.go
```diff
@@ -0,0 +1,139 @@
...
+func GetDiskDriver(diskDriver, diskPathOrBdf string) (BlockDiskDriver, error) {
+	if isBDF(diskPathOrBdf) {
+		return getDiskDriverForBDF(diskDriver, diskPathOrBdf)
+	}
+
+	return getDiskDriverForPath(diskDriver, diskPathOrBdf)
+}
+
+func getDiskDriverForBDF(diskDriver, bdf string) (BlockDiskDriver, error) {
+	executor, err := helperutil.NewExecutor(commonTypes.ProcDirectory)
```
shuo-wu, line 68: NIT: Use the executor in `go-common-lib` instead.

pkg/spdk/disk/types.go
```diff
@@ -0,0 +1,139 @@
...
+	var diskStatus DiskStatus
+	err = json.Unmarshal([]byte(output), &diskStatus)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to unmarshal disk status for BDF %s", bdf)
+	}
+
+	switch BlockDiskDriver(diskDriver) {
+	case BlockDiskDriverAuto:
+		diskPath := ""
+		if BlockDiskDriver(diskDriver) == BlockDiskDriverAuto && BlockDiskDriver(diskStatus.Driver) != UioDriverUioPciGeneric {
```
shuo-wu, line 87: `BlockDiskDriver(diskDriver) == BlockDiskDriverAuto` is duplicated

pkg/spdk/disk/types.go
```diff
@@ -0,0 +1,139 @@
...
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to get disk subsystems for %s", diskPath)
+		}
+
+		if slices.Contains(subsystems, string(BlockDiskSubsystemVirtio)) && slices.Contains(subsystems, string(BlockDiskSubsystemPci)) {
+			diskDriver := BlockDiskDriver(BlockDiskDriverVirtioBlk)
+			if slices.Contains(subsystems, string(BlockDiskSubsystemScsi)) {
+				diskDriver = BlockDiskDriver(BlockDiskDriverVirtioScsi)
+			}
+			return diskDriver, nil
```
shuo-wu, line 117: Would you explain the relationship between these drivers?

pkg/spdk/disk/types.go
```diff
@@ -0,0 +1,139 @@
...
+	err = json.Unmarshal([]byte(output), &diskStatus)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to unmarshal disk status for BDF %s", bdf)
+	}
+
+	switch BlockDiskDriver(diskDriver) {
+	case BlockDiskDriverAuto:
+		diskPath := ""
+		if BlockDiskDriver(diskDriver) == BlockDiskDriverAuto && BlockDiskDriver(diskStatus.Driver) != UioDriverUioPciGeneric {
+			devName, err := util.GetDevNameFromBDF(bdf)
```
shuo-wu, line 88: What's the difference between this `bdf` and the below `diskPath`. Can `bdf` be like `/dev/longhon/volume` while `diskPath` should be like `/dev/sdx`? And why must we use the latter one?

derekbit: BDF(Bus, Device, Function) is a unique identifier assigned to every PCIe device. Disks such as NVMe or Virtio PCI block device has an BDF like 0000:00:01.0. Users find the value by `ls -al /sys/block/`.

A diskPath is `/dev/sdx` or `/dev/nvme1n0`.

pkg/spdk/disk/types.go
derekbit: SPDK supports various types of disks, including NVMe, virtio-blk, and virtio-scsi.

NVMe disks can be managed by either NVMe bdev or AIO bdev.
VirtIO disks can be managed by virtio-blk, virtio-scsi, or AIO bdev.
To use the correct bdev, you need to identify the disk type.

Here's how to identify the disk type:

- If a block device uses the subsystems virtio and pci, it's a virtio-blk disk.
- If it uses the subsystems virtio, pci, and scsi, it's a virtio-scsi disk.

pkg/spdk/disk/types.go
derekbit: It is a wrapper of `NewNamespaceExecutor` in go-common-lib.

pkg/spdk/disk/types.go
derekbit: Fixed. Thank you.

go.mod
```diff
@@ -53,3 +53,5 @@ require (
 	k8s.io/mount-utils v0.26.0 // indirect
 	k8s.io/utils v0.0.0-20221107191617-1a15be271d1d // indirect
 )
+
+replace github.com/longhorn/go-spdk-helper v0.0.0-20240222125115-a6f538717427 => github.com/derekbit/go-spdk-helper v0.0.0-20240222164058-b8d289e3174f
```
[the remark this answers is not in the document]
derekbit: Yes. Waiting for https://github.com/longhorn/go-spdk-helper/pull/72. After it is merged, the dependency will be updated.

pkg/spdk/disk/nvme/nvme.go
```diff
@@ -0,0 +1,97 @@
...
+
+	_, err = spdksetup.Unbind(diskPath, executor)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to unbind NVMe disk %v", diskPath)
+	}
+	return true, nil
+}
+
+func (d *DiskDriverNvme) DiskGet(spdkClient *spdkclient.Client, diskName, diskPath string, timeout uint64) ([]spdktypes.BdevInfo, error) {
+	bdevs, err := spdkClient.BdevGetBdevs("", 0)
```
DamiaSan, line 76: Why not use `bdev_nvme_get_controllers`?

derekbit: We need `bdev` rather than `controller` here.

DamiaSan (dismissed):

DamiaSan (dismissed):

DamiaSan (approved):

derekbit:
ping @innobead

DamiaSan (dismissed):

pkg/spdk/disk/driver.go
```diff
@@ -0,0 +1,88 @@
...
+func DiskDelete(spdkClient *spdkclient.Client, diskName, diskPath, diskDriver string) (bool, error) {
+	driver, ok := diskDrivers[diskDriver]
+	if !ok {
+		return false, fmt.Errorf("disk driver %s is not registered", diskDriver)
+	}
+
+	return driver.DiskDelete(spdkClient, diskName, diskPath)
+}
+
+func DiskGet(spdkClient *spdkclient.Client, diskName, diskPath, diskDriver string, timeout uint64) ([]spdktypes.BdevInfo, error) {
```
shuo-wu, line 52: NIT: Can we use `UUID` as the alternative? If `UUID` is not empty, use it then verify other non-empty inputs. Otherwise, fall back to the current implementation.

pkg/util/block.go
```diff
@@ -0,0 +1,72 @@
...
+	ne, err := helperutil.NewExecutor(commonTypes.ProcDirectory)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create executor")
+	}
+
+	cmdArgs := []string{"-n", "-d", "-o", "subsystems", devPath}
+	output, err := ne.Execute(nil, "lsblk", cmdArgs, types.ExecuteTimeout)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get disk subsystems")
+	}
```
shuo-wu, line 69: We can ask `lsblk` to print out a JSON output with option `-J` then marshal it as a struct. Then:
1. The upper layer does not need to check keywords line by line.
2. We don't need so many kinds of `GetBlockDiskXXXX` functions.

pkg/spdk/disk/aio/aio.go
```diff
@@ -0,0 +1,78 @@
...
+	spdkutil "github.com/longhorn/go-spdk-helper/pkg/util"
+
+	"github.com/longhorn/longhorn-spdk-engine/pkg/spdk/disk"
+)
+
+type DiskDriverAio struct {
+}
+
+const (
+	diskDriver = "aio"
```
shuo-wu, line 21: NIT: Can we use `commonTypes.DiskDriverAio` instead? Same for others.

pkg/spdk/disk/driver.go
```diff
@@ -0,0 +1,88 @@
...
+	diskDrivers map[string]DiskDriver
+)
+
+func init() {
+	diskDrivers = make(map[string]DiskDriver)
+}
+
+func RegisterDiskDriver(diskDriver string, ops DiskDriver) error {
+	if _, exists := diskDrivers[diskDriver]; exists {
+		return fmt.Errorf("disk driver %s has already been registered", diskDriver)
```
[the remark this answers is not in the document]
derekbit: Good catch! No need to error out here.

pkg/spdk/disk/types.go
```diff
@@ -0,0 +1,139 @@
...
+	var diskStatus DiskStatus
+	err = json.Unmarshal([]byte(output), &diskStatus)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to unmarshal disk status for BDF %s", bdf)
+	}
+
+	switch BlockDiskDriver(diskDriver) {
+	case BlockDiskDriverAuto:
+		diskPath := ""
+		if BlockDiskDriver(diskStatus.Driver) != UioDriverUioPciGeneric {
```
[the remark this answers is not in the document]
derekbit: Yes, it is a disk driver that manages the disk.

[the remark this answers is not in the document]
derekbit: Rename it to `DiskDriverUioPciGeneric`

pkg/spdk/disk/driver.go
derekbit: Currently, ddiskGet calls `BdevGetBdevs` which only uses `diskName`. We don't support renaming disk for v2 data engine. Can handle the issue in another improvement ticket?

pkg/spdk/disk/aio/aio.go
derekbit: Sure! Addressd, thank you.

pkg/spdk/disk/driver.go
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/spdk/disk/types.go
```diff
@@ -0,0 +1,139 @@
...
+	output, err := spdksetup.GetDiskStatus(bdf, executor)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to get disk status for BDF %s", bdf)
+	}
+
+	var diskStatus DiskStatus
+	err = json.Unmarshal([]byte(output), &diskStatus)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to unmarshal disk status for BDF %s", bdf)
+	}
```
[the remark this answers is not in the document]
derekbit: Sure!  Addressed, thank you.

pkg/util/block.go
derekbit: It is interesting.
I tried to address the issue, but I found `subsystems` is not printed in the output of `lsblk -n -d -J <path>`.
When I added `lsblk -n -d -J -o subsystems <path>`, the results only contains `subsystems`, like
```
# lsblk -n -d -J -o subsystems /dev/nvme0n1
{
   "blockdevices": [
      {
         "subsystems": "block:nvme:pci"
      }
   ]
}
```
That means we still need to parse the results.

Currently, we only need to get subsystems (GetBlockDiskType is removed. Only GetBlockDiskSubsystem is used). Can we improve it later if we want to get more data?

pkg/spdk/disk/driver.go
DamiaSan: The RPC `bdev_get_bdevs`, which is used by `BdevGetBdevs`, can be called both with `bdev name` and `bdev alias`.
For logical volume, the `bdev name` is the lvol's uuid and the `bdev alias` is the lvol's name.
For other bdevs is the opposite, for example for a malloc bdev the `bdev name` is the malloc's uuid and the `bdev alias` is the malloc's name.

pkg/util/block.go
shuo-wu: `lsblk -O -J` should print out all fields. You can ignore marshaling unnecessary fields now:
```
lsblk -O -J /dev/sdb
{
   "blockdevices": [
      {
         "name": "sdb",
         "kname": "sdb",
         "path": "/dev/sdb",
         "maj:min": "8:16",
         "fsavail": null,
         "fssize": null,
         "fstype": null,
         "fsused": null,
         "fsuse%": null,
         "fsroots": [
             null
         ],
         "fsver": null,
         "mountpoint": null,
         "mountpoints": [
             null
         ],
         "label": null,
         "uuid": null,
         "ptuuid": null,
         "pttype": null,
         "parttype": null,
         "parttypename": null,
         "partlabel": null,
         "partuuid": null,
         "partflags": null,
         "ra": 128,
         "ro": false,
         "rm": false,
         "hotplug": false,
         "model": "Micron_5200_MTFD",
         "serial": "2005264A2CC6",
         "size": "447.1G",
         "state": "running",
         "owner": "root",
         "group": "disk",
         "mode": "brw-rw----",
         "alignment": 0,
         "min-io": 4096,
         "opt-io": 0,
         "phy-sec": 4096,
         "log-sec": 512,
         "rota": false,
         "sched": "mq-deadline",
         "rq-size": 64,
         "type": "disk",
         "disc-aln": 0,
         "disc-gran": "4K",
         "disc-max": "2G",
         "disc-zero": false,
         "wsame": "0B",
         "wwn": "0x500a0751264a2cc6",
         "rand": false,
         "pkname": null,
         "hctl": "1:0:0:0",
         "tran": "sata",
         "subsystems": "block:scsi:pci",
         "rev": "U020",
         "vendor": "ATA     ",
         "zoned": "none",
         "dax": false
      }
   ]
}
```

pkg/spdk/disk/driver.go
shuo-wu: As Damiano mentioned, it's doable. But YES I can handle this with a separate ticket.

pkg/util/block.go
derekbit: Addressed. Thank you.

pkg/spdk/disk/driver.go
derekbit: Let's handle it in a separate ticket.
But what's the benefit of UUID in the function?

derekbit: > The RPC `bdev_get_bdevs`, which is used by `BdevGetBdevs`, can be called both with `bdev name` and `bdev alias`. For logical volume, the `bdev name` is the lvol's uuid and the `bdev alias` is the lvol's name. For other bdevs is the opposite, for example for a malloc bdev the `bdev name` is the malloc's uuid and the `bdev alias` is the malloc's name.

The naming rules in spdk are pretty messy...

shuo-wu: Yes. And I tried to distinguish them in our repos:
`lvolName`: The latter part of the lvol alias. It does not contain the lvs name.
`lvolAlias`: `<lvs name>/<lvol name>`

And there are more potential conventions against the other kinds of naming like Longhorn snapshot name, snasphot lvol name, backup, and so on. If required, we can add some comments or write a doc for the related repos.

shuo-wu: > what's the benefit of UUID in the function
It's more precise. And it's convention for debugging purpose.

pkg/spdk/disk/types.go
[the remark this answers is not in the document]
shuo-wu: `UioPciGeneric` is more like a driver SPDK relies on rather than a disk driver type SPDK provides. Can we put them together?

shuo-wu (dismissed):
In general LGTM

pkg/spdk/disk/types.go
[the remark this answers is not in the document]
derekbit: I think yes, because the driver means a disk replies on which driver. For spdk, disk operations can rely on UioPciGeneric or other uio drivers.

DamiaSan (approved):

PR #110, opened 2024-02-21
test: fix test cases due to the change of namespaced Execute()

derekbit:

Issue longhorn/longhorn#7939

### DIFF
--- a/pkg/spdk_test.go
+++ b/pkg/spdk_test.go
@@ -63,7 +63,7 @@ func GetSPDKDir() string {
 	return filepath.Join(os.Getenv("GOPATH"), "src/github.com/longhorn/spdk")
 }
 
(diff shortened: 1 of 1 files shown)


innobead (approved):

innobead:
@derekbit will the change also break the 1.6 code which using ns executor?

innobead:
@derekbit Please help fix the codespell err in another PR. Thanks.

derekbit:
> @derekbit will the change also break the 1.6 code which using ns executor if we update the dependency in 1.6.

Not a breaking change. Just need to update the arguments.

derekbit:
> @derekbit Please help fix the codespell err in another PR. Thanks.

The typo is due to
```
Error: ./proto/vendor/protobuf/src/google/protobuf/empty.proto:43: re-use ==> reuse
```

Do you mean fix the proto file or fix the codespell configuration?

PR #116, opened 2024-02-24
Implement volume resizing

AaronDewes:

https://github.com/longhorn/longhorn/issues/8022

Currently, SPDK volumes can not be resized. This aims at fixing this, and allows both growing and shrinking volumes.

I've not contributed to Longhorn before and this is completely untested. I'm trying to implement & test this feature, this is just a draft.


AaronDewes:
The approach this takes is that individual replicas are all resized. A RAID device in spdk should be automatically resized if all underlying devices grow (https://github.com/longhorn/spdk/blob/6925ca6543e4ad5a5e3a5db09217ae16bf939831/module/bdev/raid/bdev_raid.c#L2102).

derekbit:
There is a dm linear device on the top of a device created by SPDK, so need to handle the dm device when expanding a volume.

DamiaSan:
I didn't know that Aaron had already done this development, in the feature request ticket https://github.com/longhorn/longhorn/issues/8022 there isn't any reference to this PR, so I understood that the development was to be done, as I actually did : https://github.com/longhorn/longhorn/issues/8022#issuecomment-2449730486.
At this point I prefer to proceed with my development on `longhorn-spdk-engine` and discard Aaron's one.
cc @derekbit

DamiaSan:
> You can proceed with your development, but the most important feature is delta rebuilding. Please co-work with @shuo-wu first.

Yes I know, I only wanted to say that the apparent issue is solved and so the development can be completed without having other blocking tasks.

For delta rebuilding, I am actually creating into `go-spdk-helper` the new API for delta bitmap and for snapshot checksum. What about snapshot checksum in `SPDK`, I have to make a change to the [PR](https://github.com/longhorn/spdk/pull/35) based on a suggestion from Shuo.
What about the work on `longhorn-spdk-engine`, I don't know the status of the developments.

> If you don't have time for the volume expansion feature (best effort in v1.8.0), feel free to transfer to @c3y1huang. Thank you.

As mentioned in the comment linked above https://github.com/longhorn/longhorn/issues/8022#issuecomment-2449730486, the work on `types`, `go-spdk-helper` and `longhorn-spdk-engine` has alredy been done and it is ready for the PR. What must be completed is the fix into SPDK, I have to finalize it and I don't think I can transfer it to Chin-Ya.

PR #118, opened 2024-03-01
Test adjustment

DamiaSan:

Issue #7578 , maybe #7559 #7810 #7697 #7693

With the introduction of lvol get xattr, the test need a greater timeout to be completed.
Moreover, the test exposed a strange behavior of the system when performing a global `sync` before to make a snapshot. No flush command arrive to spdk and I/O errors start to happen in nvme kernel driver. We encounter also frequently disconnections/reconnections between nvme driver and spdk; with kernel version older thank 6.7, all of this lead to memory corruption inside nvme driver and so to system instability. 
However, I have to check better the reason of this strange behavior, but surely what I found is that with a `sync /dev/nvmeXnY` the problem disappear and that kernel 6.7 handle much better the errors in nvme driver.


shuo-wu (approved):

shuo-wu:
The conflict is not resolved yet. Would you update the PR again?

shuo-wu (approved):

PR #131, opened 2024-04-19
Detach nvmf controller for a remote replica while removing replica from engine

derekbit:
- **vendor: update go-spdk-helper**
- **Detach nvmf controller for a remote replica while removing replica from engine**

Issue longhorn/longhorn#8354

To avoid the repeatedly reconnection due to the existing nvmf controller, detach nvmf controller for a remote replica while removing a replica from its engine.


DamiaSan:
Could you describe better the scenario? What do you mean by removing a replica from its engine?

derekbit:
> Could you describe better the scenario? What do you mean by removing a replica from its engine?

If a node is down or an instance-manager pod is unexpectedly deleted, replicas on the node are marked as error. In that case, longhorn-manager needs to delete the replicas from their engines. These replicas can be added back to engines while rebuilding replicas or be deleted forever.

pkg/spdk/engine.go
```diff
@@ -1047,8 +1055,15 @@ func (e *Engine) ReplicaDelete(spdkClient *spdkclient.Client, replicaName, repli
...
+	e.log.Infof("Removing base bdev %v from engine", e.ReplicaBdevNameMap[replicaName])
 	if _, err := spdkClient.BdevRaidRemoveBaseBdev(e.ReplicaBdevNameMap[replicaName]); err != nil && !jsonrpc.IsJSONRPCRespErrorNoSuchDevice(err) {
-		return err
+		return errors.Wrapf(err, "failed to remove base bdev %s for deleting replica %s", e.ReplicaBdevNameMap[replicaName], replicaName)
+	}
+
+	controllerName := helperutil.GetNvmeControllerNameFromNamespaceName(replicaName)
+	e.log.Infof("Detaching controller %v", controllerName)
+	if _, err := spdkClient.BdevNvmeDetachController(controllerName); err != nil && !jsonrpc.IsJSONRPCRespErrorNoSuchDevice(err) {
+		return errors.Wrapf(err, "failed to detach controller %s for deleting replica %s", controllerName, replicaName)
```
shuo-wu, line 1066: Shouldn't we suspend the dm device and restart the rest of the frontend before doing this?

derekbit: No, we don't need to suspend the dm device.
If one of replicas is being deleted, the IO will be queued at SPDK raid layer. After deletion, the IO continues.

shuo-wu: OK, got it. But it's better to check if the replica is the local lvol before`BdevNvmeDetachController`. 
Besides, the log message should be:
```
Detaching the corresponding NVMf controller %v during remote replica %s delete
```

derekbit: Addressed. Thank you.

derekbit:
@DamiaSan  @shuo-wu  Ready to review again. Thank you.

shuo-wu (approved):

PR #158, opened 2024-06-21
ci: simplify spdk_tgt start and stop

derekbit:

Issue #


innobead (approved):

shuo-wu:
I prefer to launch a new tgt_target for each test case so that we can avoid some potential leftovers left by the previous tests.

innobead:
Sounds good. Or before running each case, just cleaning up the potential leftover instances?

derekbit:
> I prefer to launch a new tgt_target for each test case so that we can avoid some potential leftovers left by the previous tests.

Got it. The original method launching new tgt_target leads to random timeout error in both arm64 and amd64 platforms. I think it is due to something wrong in StartTarget function. We need to fix it first.

innobead:
or we don't need to restart the target but rather use different resource names for each test so that they will be independent of each other. Although there would be a chance to have leftover instances from executed tests, the following tests will not be impacted.

PR #162, opened 2024-06-25
engine: add timestamp to user created snapshot

DamiaSan:

Issue #7641

I added the timestamp only to the snapshots created by the user, because they are the only ones listed in the UI.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -139,19 +139,23 @@ func (c *SPDKClient) ReplicaWatch(ctx context.Context) (*api.ReplicaStream, erro
 }
 
 func (c *SPDKClient) ReplicaSnapshotCreate(name, snapshotName string, opts *api.SnapshotOptions) error {
-	if name == "" || snapshotName == "" {
-		return fmt.Errorf("failed to create SPDK replica snapshot: missing required parameter name or snapshot name")
+	if name == "" || snapshotName == "" || opts == nil {
+		return fmt.Errorf("failed to create SPDK replica snapshot: missing required parameter name, snapshot name or opts")
 	}
 
 	client := c.getSPDKServiceClient()
 	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceTimeout)
 	defer cancel()
 
-	_, err := client.ReplicaSnapshotCreate(ctx, &spdkrpc.SnapshotRequest{
-		Name:         name,
-		SnapshotName: snapshotName,
-		UserCreated:  opts != nil && opts.UserCreated,
-	})
+	snapshotRequest := spdkrpc.SnapshotRequest{
+		Name:             name,
+		SnapshotName:     snapshotName,
+		UserCreated:      opts.UserCreated,
+		SnaphotTimestamp: opts.Timestamp,
+	}
+
+	_, err := client.ReplicaSnapshotCreate(ctx, &snapshotRequest)
+
 	return errors.Wrapf(err, "failed to create SPDK replica %s snapshot %s", name, snapshotName)
 }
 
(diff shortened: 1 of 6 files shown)


pkg/client/client.go
```diff
@@ -445,11 +456,16 @@ func (c *SPDKClient) EngineSnapshotCreate(name, snapshotName string, opts *api.S
...
 
-	resp, err := client.EngineSnapshotCreate(ctx, &spdkrpc.SnapshotRequest{
+	snapshotRequest := spdkrpc.SnapshotRequest{
 		Name:         name,
 		SnapshotName: snapshotName,
 		UserCreated:  opts != nil && opts.UserCreated,
-	})
+	}
+	if opts != nil {
+		snapshotRequest.SnaphotTimestamp = opts.Timestamp
```
derekbit, line 465: Same as above

pkg/spdk/server.go
```diff
@@ -452,7 +452,12 @@ func (s *Server) ReplicaSnapshotCreate(ctx context.Context, req *spdkrpc.Snapsho
 		return nil, grpcstatus.Errorf(grpccodes.NotFound, "cannot find replica %s during snapshot create", req.Name)
 	}
 
-	return r.SnapshotCreate(spdkClient, req.SnapshotName, &api.SnapshotOptions{UserCreated: req.UserCreated})
+	opts := api.SnapshotOptions{
+		UserCreated: req.UserCreated,
+		Timestamp:   req.SnaphotTimestamp,
+	}
+
+	return r.SnapshotCreate(spdkClient, req.SnapshotName, &opts)
```
derekbit, line 460: Why not
```suggestion
	opts := &api.SnapshotOptions{
		UserCreated: req.UserCreated,
		Timestamp:   req.SnaphotTimestamp,
	}

	return r.SnapshotCreate(spdkClient, req.SnapshotName, opts)
```

pkg/spdk/server.go
```diff
@@ -660,7 +665,12 @@ func (s *Server) ReplicaRebuildingDstSnapshotCreate(ctx context.Context, req *sp
 		return nil, grpcstatus.Errorf(grpccodes.NotFound, "cannot find replica %s during rebuilding dst snapshot create", req.Name)
 	}
 
-	if err = r.RebuildingDstSnapshotCreate(spdkClient, req.SnapshotName, &api.SnapshotOptions{UserCreated: req.UserCreated}); err != nil {
+	opts := api.SnapshotOptions{
+		UserCreated: req.UserCreated,
+		Timestamp:   req.SnaphotTimestamp,
+	}
+
+	if err = r.RebuildingDstSnapshotCreate(spdkClient, req.SnapshotName, &opts); err != nil {
```
derekbit, line 673: Same as above.

pkg/client/client.go
```diff
@@ -147,11 +147,17 @@ func (c *SPDKClient) ReplicaSnapshotCreate(name, snapshotName string, opts *api.
 	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceTimeout)
 	defer cancel()
 
-	_, err := client.ReplicaSnapshotCreate(ctx, &spdkrpc.SnapshotRequest{
+	snapshotRequest := spdkrpc.SnapshotRequest{
 		Name:         name,
 		SnapshotName: snapshotName,
 		UserCreated:  opts != nil && opts.UserCreated,
```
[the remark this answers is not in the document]
DamiaSan: Good improvement, thanks

pkg/client/client.go
DamiaSan: Here `SnapshotTimestamp` is not needed, I removed it. Actually, not even `UserCreated` is needed, so the `opts` parameter is unnecessary, maybe it is better to remove it after this PR will be merged. WDYT?

pkg/spdk/engine.go
```diff
@@ -772,7 +772,10 @@ func (e *Engine) ReplicaAddStart(spdkClient *spdkclient.Client, replicaName, rep
 
 	// TODO: For online rebuilding, the IO should be paused first
 	snapshotName := GenerateRebuildingSnapshotName()
-	updateRequired, err = e.snapshotOperationWithoutLock(spdkClient, replicaClients, snapshotName, SnapshotOperationCreate, nil)
+	opts := api.SnapshotOptions{
+		Timestamp: time.Now().String(),
+	}
+	updateRequired, err = e.snapshotOperationWithoutLock(spdkClient, replicaClients, snapshotName, SnapshotOperationCreate, &opts)
```
derekbit, line 778: nit:
```suggestion
	opts := &api.SnapshotOptions{
		Timestamp: time.Now().String(),
	}
	updateRequired, err = e.snapshotOperationWithoutLock(spdkClient, replicaClients, snapshotName, SnapshotOperationCreate, opts)
```

pkg/client/client.go
derekbit: > not even UserCreated is needed

Can you elaborate more on it?

DamiaSan: Both user created and snapshot timestamp flags are internally resolved in `longhorn-spdk-engine`, so it is not needed to insert them as parameters in the function `EngineSnapshotCreate`, which is called by `longhorn-instance-manager`.

derekbit: I see. Then, we can remove the parameter `opts`.

derekbit: Why not remove it in this PR?

derekbit (approved):
LGTM

PR #167, opened 2024-06-27
feat(replica): handle new shallow copy async APIs

DamiaSan:

Issue #8820

This is a tentative PR. tell me if if the actual implementation can be fine.


pkg/spdk/replica.go
```diff
@@ -1230,22 +1230,46 @@ func (r *Replica) RebuildingSrcDetach(spdkClient *spdkclient.Client, dstReplicaN
...
+
+func (r *Replica) SnapshotCheckShallowCopy(spdkClient *spdkclient.Client, operationId uint32) (bool, error) {
+	var status *spdktypes.ShallowCopyStatus
+	var err error
+
+	if status, err = spdkClient.BdevLvolCheckShallowCopy(operationId); err != nil {
+		return false, err
+	}
+
+	if status.State == "error" {
```
derekbit, line 1259: Define the constants 
```
type ShallowCopyState string

ShallowCopyStateError = ShallowCopyState("error")
ShallowCopyStateComplete = ShallowCopyState("complete")
...
```

DamiaSan: If I define the type `ShallowCopyState` I get a mismatch error when comparing with a string. Moreover, what is the benefit to define such a type? I have simply defined 2 const string. Is it ok?

derekbit: It is just a convention for state management in Longhorn and helps developers understand what states we have.

> If I define the type ShallowCopyState I get a mismatch error when comparing with a string
> I have simply defined 2 const string. Is it ok?

Can you try
```
if ShallowCopyState(status.State) == ShallowCopyStateError {
    ...
}
```

pkg/spdk/replica.go
```diff
@@ -31,8 +31,13 @@ import (
...
 )
 
+type ShallowCopyState string
+
 const (
 	restorePeriodicRefreshInterval = 2 * time.Second
+
+	ShallowCopyStateError    = ShallowCopyState("error")
+	ShallowCopyStateComplete = ShallowCopyState("complete")
 )
```
derekbit, line 42: Move to pkg/spdk/types.go

derekbit (approved):

PR #177, opened 2024-07-11
fix(snapshot): revert "engine: sync only own device before snapshot"

derekbit:

Issue Longhorn/longhorn#8977

The commit (da2a02e) results in a regression that v2 volume data are not sync before taking a snapshot. Using sync directly with a block device path does not work as expected.

According to the doc of the sync command, what it does is to synchronize the filesystem which contains the specified file.

### DIFF
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -1261,7 +1261,8 @@ func (e *Engine) snapshotOperation(spdkClient *spdkclient.Client, inputSnapshotN
 				e.log.WithError(err).Errorf("WARNING: failed to get the executor for snapshot op %v with snapshot %s, will skip the sync and continue", snapshotOp, inputSnapshotName)
 			} else {
 				e.log.Infof("Requesting system sync %v before snapshot", devicePath)
-				if _, err := ne.Execute(nil, "sync", []string{devicePath}, SyncTimeout); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/spdk/engine.go
```diff
@@ -1261,7 +1261,7 @@ func (e *Engine) snapshotOperation(spdkClient *spdkclient.Client, inputSnapshotN
 				e.log.WithError(err).Errorf("WARNING: failed to get the executor for snapshot op %v with snapshot %s, will skip the sync and continue", snapshotOp, inputSnapshotName)
 			} else {
 				e.log.Infof("Requesting system sync %v before snapshot", devicePath)
-				if _, err := ne.Execute(nil, "sync", []string{devicePath}, SyncTimeout); err != nil {
+				if _, err := ne.Execute(nil, "sync", []string{}, SyncTimeout); err != nil {
```
c3y1huang, line 1264: Should we also revert the comment and add the finding to it?
https://github.com/longhorn/longhorn-spdk-engine/commit/da2a02e86462fb4f3abac4d0f91f2380ff34053c#diff-2f9990d3c794aaa008aacf70ecd9ae72fc65febf9eee3ea038463cc542e43c36L1100

derekbit: Yes, revert is better. Updated. Thank you.

PR #179, opened 2024-07-16
Online Rebuilding

shuo-wu:

Issue Longhorn/longhorn#7199

Will implement the control layer then.


DamiaSan:
I have seen in many point that you make distinction between local or remote lvols. In the first case you use directly the lvols, otherwise you use nvme-of exported ones. But if I remember correctly, the implementation of live upgrade made by @derekbit needs to treat all the lvols as exported nvme-of ones, also if they are local. Can this in some way cause some kind of conflicts or maybe semplify the code?

shuo-wu:
> But if I remember correctly, the implementation of live upgrade made by @derekbit needs to treat all the lvols as exported nvme-of ones, also if they are local. Can this in some way cause some kind of conflicts or maybe semplify the code?

The latest implementation no longer makes many differences between remote and local replicas. 
The main exception is how we clone a rebuilding lvol for the dst replica from the external snapshot. Do we still need to blindly launch a nvmf layer for the src replica exposed snapshot no matter if src and dst are on the same host?

derekbit:
> > But if I remember correctly, the implementation of live upgrade made by @derekbit needs to treat all the lvols as exported nvme-of ones, also if they are local. Can this in some way cause some kind of conflicts or maybe semplify the code?
> 
> The latest implementation no longer makes many differences between remote and local replicas. The main exception is how we clone a rebuilding lvol for the dst replica from the external snapshot. Do we still need to blindly launch a nvmf layer for the src replica exposed snapshot no matter if src and dst are on the same host?

The only concern is the base based added in a raid.
The dest bdev in the implemeation is exposed and attached. LGTM. 
```
  dstHeadLvolBdevName, err := e.connectReplica(spdkClient, dstReplicaName, dstHeadLvolAddress)
	  if err != nil {
		  return "", nil, err
	  }
	  if _, err := spdkClient.BdevRaidGrowBaseBdev(e.Name, dstHeadLvolBdevName); err != nil {
		  engineErr = err
		  return "", nil, errors.Wrapf(err, "failed to adding the rebuilding replica %s head bdev %s to the base bdev list for engine %s", dstReplicaName, dstHeadLvolBdevName, e.Name)
```

For the clone part, it is ok.

pkg/spdk/replica.go
```diff
@@ -1202,77 +1268,62 @@ func (r *Replica) RebuildingSrcAttach(spdkClient *spdkclient.Client, dstReplicaN
...
+// RebuildingSrcShallowCopyCheck asks the src replica to check the shallow copy progress and status via the shallow copy op ID returned by RebuildingSrcShallowCopyStart.
+func (r *Replica) RebuildingSrcShallowCopyCheck(spdkClient *spdkclient.Client, shallowCopyOpID uint32) (status *spdktypes.ShallowCopyStatus, err error) {
+	return spdkClient.BdevLvolCheckShallowCopy(shallowCopyOpID)
 }
 
-func (r *Replica) RebuildingDstStart(spdkClient *spdkclient.Client, exposeRequired bool) (address string, err error) {
+// RebuildingDstStart asks the dst replica to create a new head lvol based on the external snapshot of the src replica and blindly expose it as a NVMf bdev.
+// It returns the new head lvol address <IP>:<Port>.
+// Notice that input `externalSnapshotAddress` is the alias of the external snapshot lvol if src and dst have on the same IP, otherwise it's the NVMf address of the external snapshot lvol.
+func (r *Replica) RebuildingDstStart(spdkClient *spdkclient.Client, srcReplicaName, srcReplicaAddress, externalSnapshotName, externalSnapshotAddress string, rebuildingSnapshotList []*api.Lvol) (address string, err error) {
```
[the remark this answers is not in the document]
DamiaSan: Yes, it is. We use external snapshot to avoid the use of write only volume.

DamiaSan:
> > But if I remember correctly, the implementation of live upgrade made by @derekbit needs to treat all the lvols as exported nvme-of ones, also if they are local. Can this in some way cause some kind of conflicts or maybe semplify the code?
> 
> The latest implementation no longer makes many differences between remote and local replicas. The main exception is how we clone a rebuilding lvol for the dst replica from the external snapshot. Do we still need to blindly launch a nvmf layer for the src replica exposed snapshot no matter if src and dst are on the same host?

Just a question: suppose that a node completely loose all the data on its disk, so that you have to recreate all volumes that were on it. What about the raid bdev that were on that node, are they recreated on another node?
What I want to understand is that, when you create the clone of the external snapshot, if there is the possibility that the raid is on the same node and so that you add this clone directly to the raid and not via nvmf export/attach.
As Derek said, the only important thing is that the base bdevs of the raid are always nvmf exported/attached lvols.

shuo-wu:
> What I want to understand is that, when you create the clone of the external snapshot, if there is the possibility that the raid is on the same node and so that you add this clone directly to the raid and not via nvmf export/attach.

No. The clone of the external snapshot, which is the head lvol of the rebuilding dst replica, will be always exposed as a nvmf bdev for RAID usage.

pkg/spdk/replica.go
```diff
@@ -1092,39 +1128,58 @@ func (r *Replica) RebuildingSrcStart(spdkClient *spdkclient.Client, localReplica
...
+		return "", fmt.Errorf("cannot find snapshot %s for for replica %s rebuilding src start", exposedSnapshotName, r.Name)
 	}
-	// TODO: After launching online rebuilding, the destination lvol name would be GetReplicaRebuildingLvolName(dstReplicaName)
-	dstRebuildingLvolName := dstReplicaName
 
-	if dstRebuildingLvolIP == r.IP {
-		dstReplicaLvsName := localReplicaLvsNameMap[dstReplicaName]
-		if dstReplicaLvsName == "" {
-			return fmt.Errorf("cannot find dst replica %s from the local replica map for replica %s rebuilding src start", dstReplicaName, r.Name)
+	isSrcDstOnSameHost := r.IP == strings.Split(dstReplicaAddress, ":")[0]
```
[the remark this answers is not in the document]
shuo-wu: It's possible. As we discussed yesterday, it won't affect anything. We can re-visit this part later.

pkg/spdk/replica.go
```diff
@@ -1092,39 +1128,58 @@ func (r *Replica) RebuildingSrcStart(spdkClient *spdkclient.Client, localReplica
...
+	}
+
+	if !isSrcDstOnSameHost {
+		r.rebuildingSrcCache.exposedSnapshotPort, _, err = r.portAllocator.AllocateRange(1)
+		if err != nil {
+			return "", err
+		}
+		nguid := commonUtils.RandomID(nvmeNguidLength)
+		if err := spdkClient.StartExposeBdev(helpertypes.GetNQN(snapLvol.Name), snapLvol.UUID, nguid, r.IP, strconv.Itoa(int(r.rebuildingSrcCache.exposedSnapshotPort))); err != nil {
+			return "", err
```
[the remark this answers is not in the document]
shuo-wu: Theoretically, no matter whether rebuilding succeeds or not, the engine should call `srcReplicaServiceCli.ReplicaRebuildingSrcFinish` to clean up all rebuilding-related stuff. 
But here we haven't handled these parts perfectly. For now I just add a TODO comment there.

pkg/spdk/replica.go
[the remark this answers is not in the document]
shuo-wu: As we mentioned, `external snapshot` means a snapshot does not belong to this replica. It's shared from the src replica.

derekbit:
The overall flow of the online rebuilding
```
|-> ReplicaAddStart(dstReplicaName, dstReplicaAddress)
    |-> Find a src replica from e.ReplicaAddressMap
    |-> Create a spdk client for the src replica
    |-> Suspend IO
    |-> Create a snapshot for the all replicas
    |-> Get a snapshot rebuilding list
    |-> ReplicaRebuildingSrcStart(): Ask the source replica to expose the newly created snapshot if the source replica and destination replica are not on the same node. Get the external snapshot lvol address from this function call
        |-> RebuildingSrcStart(dstReplicaName, dstReplicaAddress, exposedSnapshotName): expose the parent snapshot of the head.

            example:
                head -> snap1 -> snap2 -> ... -> snapN, where snap1 is exposedSnapshotName
      
    |-> ReplicaRebuildingDstStart(): The destination replica attaches the source replica exposed snapshot as the external snapshot then create a head based on it.
        |-> RebuildingDstStart()
            |-> Attach the external snapshot exposed by the source replica
            |-> Creates a logical volume based on an external snapshot bdev or an local snapshot lvol
            |-> Expose the cloned logical volume
    |-> Connect dest replica (attach)
    |-> Add the attached dest bdev to the raid bdev by BdevRaidGrowBaseBdev()
    |-> Resume IO
|-> goroutine
    |-> ReplicaShallowCopy():
        |-> Rebuilding snapshot one by one with a DFS way
    |-> ReplicaAddFinish
        |-> Suspend IO
        |-> ReplicaRebuildingDstFinish(): Change the parent of the head and do cleanup
        |-> ReplicaRebuildingSrcFinish(): The source replica will stop exposing the snapshot and wipe the rebuilding info.
        |-> Resume IO

```

pkg/spdk/engine.go
```diff
@@ -1128,40 +1131,118 @@ func (e *Engine) ReplicaShallowCopy(dstReplicaName, dstReplicaAddress string) (e
...
-	// TODO: Connect the previously found latest snapshot lvol with the head lvol chain
-	// TODO: For online rebuilding, the rebuilding lvol must be unexposed
-	if err = dstReplicaServiceCli.ReplicaRebuildingDstFinish(dstReplicaName, e.IP == dstReplicaIP); err != nil {
+	// The source replica will stop exposing the snapshot and wipe the rebuilding info.
+	if err = srcReplicaServiceCli.ReplicaRebuildingSrcFinish(srcReplicaName, dstReplicaName); err != nil {
+		// TODO: Should we mark this healthy replica as error?
 		return err
 	}
 
+	e.ReplicaModeMap[dstReplicaName] = types.ModeRW
```
derekbit, line 1233: As discussed with @shuo-wu, the mode cannot take effect on the base bdev of the RAID.
We need to improve it and ensure that the IO mode of a base bdev is consistent with the replica mode map in the future.
cc @DamiaSan

pkg/spdk/engine.go
```diff
@@ -1128,40 +1131,118 @@ func (e *Engine) ReplicaShallowCopy(dstReplicaName, dstReplicaAddress string) (e
...
+	srcReplicaServiceCli, err := GetServiceClient(srcReplicaAddress)
+	if err != nil {
+		return err
+	}
+	dstReplicaServiceCli, err := GetServiceClient(dstReplicaAddress)
+	if err != nil {
+		return err
+	}
+
+	var engineErr error
```
derekbit, line 1178: Can you explain the purpose of `err` and `engineErr`? Suggest adding a comment to prevent their misuse.

pkg/spdk/engine.go
```diff
@@ -1128,40 +1131,118 @@ func (e *Engine) ReplicaShallowCopy(dstReplicaName, dstReplicaAddress string) (e
...
+				e.log.WithError(err).Errorf("Failed to finish rebuilding dst replica %s with address %s from src replica %s with address %s, will mark the rebuilding replica mode as ERR", dstReplicaName, dstReplicaAddress, srcReplicaName, srcReplicaAddress)
+				updateRequired = true
 			}
-			break
 		}
+	}()
+
+	// Pause the IO again by suspending the NVMe initiator
+	if e.Frontend == types.FrontendSPDKTCPBlockdev {
+		if err = e.initiator.Suspend(false, true); err != nil {
```
derekbit, line 1211: Does it need to flush IOs in the suspend?

derekbit: Should we set `egineErr=err` if `Suspend()` failed?

pkg/spdk/engine.go
```diff
@@ -918,129 +913,127 @@ func (e *Engine) ReplicaAddStart(spdkClient *spdkclient.Client, replicaName, rep
...
+		if replicaMode != types.ModeRW {
+			continue
+		}
+		srcReplicaName = replicaName
+		srcReplicaAddress = e.ReplicaAddressMap[replicaName]
+		break
+	}
+	if srcReplicaName == "" || srcReplicaAddress == "" {
+		return "", nil, fmt.Errorf("cannot find an RW replica in engine %s during replica %s add", e.Name, dstReplicaName)
+	}
```
derekbit, line 934: nit: abstrace the codes into a function

pkg/spdk/replica.go
```diff
@@ -1285,75 +1336,148 @@ func (r *Replica) RebuildingDstStart(spdkClient *spdkclient.Client, exposeRequir
...
+				r.rebuildingDstCache.rebuildingError = err.Error()
+			}
 		} else {
 			if r.State != types.InstanceStateError {
 				r.ErrorMsg = ""
 			}
 		}
+		if r.rebuildingDstCache.rebuildingError != "" && r.rebuildingDstCache.rebuildingState != types.ProgressStateError {
+			r.rebuildingDstCache.rebuildingState = types.ProgressStateError
+		}
```
derekbit, line 1361: Can we directly update `r.rebuildingDstCache.rebuildingState` to `types.ProgressStateError` while updating `rebuildingError`?

pkg/spdk/engine.go
```diff
@@ -918,129 +913,127 @@ func (e *Engine) ReplicaAddStart(spdkClient *spdkclient.Client, replicaName, rep
...
-			e.ReplicaModeMap[replicaName] = types.ModeRW
-			e.ReplicaBdevNameMap[replicaName] = bdevName
-			replicaBdevList = append(replicaBdevList, bdevName)
-			updateRequired = true
-		}
+	dstHeadLvolBdevName, err := e.connectReplica(spdkClient, dstReplicaName, dstHeadLvolAddress)
+	if err != nil {
+		return "", nil, err
+	}
+	if _, err := spdkClient.BdevRaidGrowBaseBdev(e.Name, dstHeadLvolBdevName); err != nil {
```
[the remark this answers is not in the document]
DamiaSan: Also if you put the base bdev in write only mode, some read operations are needed because of the copy on write: if you write on the head lvol over a cluster that is not still allocated, this cluster is copied from the base snapshot and then updated. So, in SPDK, also in you put the base bdev in write only mode, you continue to need a parent for that volume.
For this reason the development to set the base bdev as write only was discarded last year and we started to use the external snapshot, which resolve the above issue.

pkg/spdk/engine.go
shuo-wu: Add comment:
```
// engineErr will be set when the engine failed to do any non-recoverable operations, then there is no way to make the engine continue working. Typically, it's related to the frontend suspend or resume failures.
// While err means replica-related operation errors. It will fail the current replica add flow.
```

pkg/spdk/engine.go
shuo-wu: When the suspend fails in the middle, I am not sure if the engine still works. Hence I prefer to consider the engine as ERR first.

pkg/spdk/replica.go
```diff
@@ -1285,75 +1336,148 @@ func (r *Replica) RebuildingDstStart(spdkClient *spdkclient.Client, exposeRequir
...
-	// 	if err != nil {
-	// 		return "", err
-	// 	}
-	// 	if err := spdkClient.StartExposeBdev(helpertypes.GetNQN(lvolName), r.rebuildingLvol.UUID, r.IP, strconv.Itoa(int(r.rebuildingPort))); err != nil {
-	// 		return "", err
-	// 	}
-	// }
+	if r.ChainLength != 2 {
+		return "", fmt.Errorf("invalid chain length %d for dst replica %v rebuilding start", r.ChainLength, r.Name)
+	}
```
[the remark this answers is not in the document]
shuo-wu: The explanation is added in the variable definition instead.

pkg/spdk/engine.go
shuo-wu: > Does it need to flush IOs in the suspend?

The input `noflush` is set to `false` when calling `e.initiator.Suspend(false, true)`. I would assume the initiator can handle it automatically.

pkg/spdk/engine.go
shuo-wu: Let's add TODO and explanation comments for this first.

pkg/spdk/replica.go
```diff
@@ -1285,75 +1336,148 @@ func (r *Replica) RebuildingDstStart(spdkClient *spdkclient.Client, exposeRequir
...
+		nvmeBdevNameList, err := spdkClient.BdevNvmeAttachController(externalSnapshotLvolName, helpertypes.GetNQN(externalSnapshotLvolName),
+			externalExternalSnapshotIP, externalExternalSnapshotPort, spdktypes.NvmeTransportTypeTCP, spdktypes.NvmeAddressFamilyIPv4,
+			helpertypes.DefaultCtrlrLossTimeoutSec, helpertypes.DefaultReconnectDelaySec, helpertypes.DefaultFastIOFailTimeoutSec,
+			helpertypes.DefaultMultipath)
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to attach the external src snapshot %s with address %s as a NVMe bdev for dst replica %v rebuilding start", externalSnapshotLvolName, externalSnapshotAddress, r.Name)
+		}
+		if len(nvmeBdevNameList) != 1 {
+			return "", fmt.Errorf("got zero or multiple results when attaching the external src snapshot %s with address %s as a NVMe bdev: %+v", externalSnapshotLvolName, externalSnapshotAddress, nvmeBdevNameList)
+		}
```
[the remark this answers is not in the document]
shuo-wu: Yes. Refactored both engine and replica codes with util function `connectNVMfBdev` and `disconnectNVMfBdev`

DamiaSan (dismissed):
LGTM

pkg/spdk/engine.go
```diff
@@ -49,6 +49,7 @@ type Engine struct {
 	Nqn                string
 	Nguid              string
 
+	initiator    *nvme.Initiator
```
derekbit, line 52: Is caching the `initiator` just to avoid reloading the device?

derekbit (approved):
In general, LGTM. Just few comments.

pkg/client/client.go
```diff
@@ -580,7 +676,7 @@ func (c *SPDKClient) EngineReplicaAdd(engineName, replicaName, replicaAddress st
 	}
 
 	client := c.getSPDKServiceClient()
-	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceLongTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceTimeout)
```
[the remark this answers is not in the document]
shuo-wu: I am not sure... For now we can use the default timeout. 
During the preparation stage, there will be 3 calls: snapshot creation calls to all replicas, a src replica preparation call, and a dst replica preparation call. Besides, there is a frontend suspend and resume call.

pkg/spdk/replica.go
```diff
@@ -1285,75 +1271,137 @@ func (r *Replica) RebuildingDstStart(spdkClient *spdkclient.Client, exposeRequir
...
 			r.ErrorMsg = err.Error()
+			if r.rebuildingDstCache.rebuildingError == "" {
+				r.rebuildingDstCache.rebuildingError = err.Error()
+			}
 		} else {
 			if r.State != types.InstanceStateError {
 				r.ErrorMsg = ""
 			}
 		}
+		if r.rebuildingDstCache.rebuildingError != "" && r.rebuildingDstCache.rebuildingState != types.ProgressStateError {
```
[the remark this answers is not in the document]
shuo-wu: Switch to set `rebuildingError` and `rebuildingState` simultaneously.

pkg/spdk/engine.go
shuo-wu: I see. There is no need to flush the IO here

pkg/spdk/engine.go
```diff
@@ -1128,42 +1060,124 @@ func (e *Engine) ReplicaShallowCopy(dstReplicaName, dstReplicaAddress string) (e
...
+			e.UpdateCh <- nil
+		}
+	}()
+
+	defer func() {
+		if err != nil {
+			if e.State != types.InstanceStateError {
+				e.State = types.InstanceStateError
+				updateRequired = true
 			}
```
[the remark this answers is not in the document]
shuo-wu: The frontend suspend/resume calls may return an error.

pkg/spdk/replica.go
```diff
@@ -1137,142 +1149,116 @@ func (r *Replica) RebuildingSrcFinish(spdkClient *spdkclient.Client, dstReplicaN
 		}
 	}()
 
-	if r.rebuildingDstReplicaName == "" && r.rebuildingDstBdevName == "" && r.rebuildingDstBdevType == "" {
-		return nil
+	if r.rebuildingSrcCache.dstReplicaName != "" && r.rebuildingSrcCache.dstReplicaName != dstReplicaName {
+		return fmt.Errorf("found mismatching between the required dst replica name %s and the recorded dst replica name %s for replica %s rebuilding src finish", dstReplicaName, r.rebuildingSrcCache.dstReplicaName, r.Name)
 	}
```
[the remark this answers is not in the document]
shuo-wu: I am not sure. Probably someone wrongly calls this API with invalid parameters. Do you think we can remove it for now?

pkg/spdk/replica.go
```diff
@@ -1371,76 +1419,268 @@ func (r *Replica) RebuildingDstFinish(spdkClient *spdkclient.Client, unexposeReq
...
+	if r.rebuildingDstCache.rebuildingState == types.ProgressStateInProgress &&
+		r.rebuildingDstCache.processingState != types.ProgressStateComplete && r.rebuildingDstCache.processingState != types.ProgressStateError {
+		srcReplicaServiceCli, err := GetServiceClient(r.rebuildingDstCache.srcReplicaAddress)
+		if err != nil {
+			return nil, err
+		}
+		state, copiedClusters, totalClusters, errorMsg, err := srcReplicaServiceCli.ReplicaRebuildingSrcShallowCopyCheck(r.rebuildingDstCache.srcReplicaName, r.Name, r.rebuildingDstCache.processingSnapshotName, r.rebuildingDstCache.processingOpID)
+		if err != nil {
+			errorMsg = errors.Wrapf(err, "dst replica %s failed to check the shallow copy status from src replica %s for snapshot %s", r.Name, r.rebuildingDstCache.srcReplicaName, r.rebuildingDstCache.processingSnapshotName).Error()
+		}
```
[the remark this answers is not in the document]
shuo-wu: I see. When `err != nil", we will suddenly get 0 `copiedClusters`, which will overwrite the previously non-zero value.

pkg/spdk/engine.go
shuo-wu: Yes. Launching an `initiator` is trivial. And it's frequently used in `engine`

pkg/client/client.go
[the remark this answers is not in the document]
derekbit: I see. Let' keep it as it is.

derekbit:
@shuo-wu The PR is still a DRAFT. Will you convert it to READY?

pkg/spdk/replica.go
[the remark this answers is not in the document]
derekbit: No, just curious, we can retain the codes.

derekbit (dismissed):
The changes LGTM. Thank you.

pkg/spdk/engine.go
```diff
@@ -1128,42 +1060,124 @@ func (e *Engine) ReplicaShallowCopy(dstReplicaName, dstReplicaAddress string) (e
...
+				e.ErrorMsg = ""
 			}
-			break
 		}
+	}()
+
+	// Blindly ask the source replica to detach the rebuilding lvol
+	if srcReplicaErr := srcReplicaServiceCli.ReplicaRebuildingSrcDetach(srcReplicaName, dstReplicaName); srcReplicaErr != nil {
+		e.log.WithError(srcReplicaErr).Errorf("Engine failed to detach the rebuilding lvol for rebuilding src replica %s, will ignore this error and continue", srcReplicaName)
 	}
```
derekbit, line 1099: Will there be any orphaned attached rebuilding lvol if failed to detach?

shuo-wu: Theoretically YES. But we have not idea to handle it if this detachment fails. And we cannot make it block the flow

derekbit: I see. I think we can figure out a solution to it.
Can you leave a comment for it? Thank you.

derekbit (approved):

PR #186, opened 2024-07-25
Implement snapshot purge

shuo-wu:

Issue longhorn/longhorn#7199

Need to rely on purge APIs to clean up system snapshots created by rebuilding automatically


pkg/spdk/replica.go
```diff
@@ -1071,6 +1071,51 @@ func (r *Replica) SnapshotRevert(spdkClient *spdkclient.Client, snapshotName str
...
+
+	if r.ChainLength < 2 {
+		return fmt.Errorf("invalid chain length %d for replica snapshot purge", r.ChainLength)
+	}
+
+	// delete all non-user-created snapshots
+	for snapshotLvolName, snapSvcLvol := range r.SnapshotLvolMap {
+		if snapSvcLvol.UserCreated {
+			continue
+		}
```
derekbit, line 1102: We have a [bug](https://github.com/longhorn/longhorn/issues/9054#issuecomment-2245049025) in v1.6.2.
The `userCreated` of snapshots for v2 volumes created in v1.6.2 are set to `false`.
The purge will lead to these snapshots being deleted.
We can add this in the `important note`? Is it OK? cc @shuo-wu @DamiaSan @innobead

shuo-wu: YES. I am fine with it.

pkg/types/types.go
```diff
@@ -71,4 +71,7 @@ const (
 	ProgressStateComplete   = "complete"
 	ProgressStateInProgress = "in_progress"
 	ProgressStateStarting   = "starting"
+
+	// SPDKShallowCopyStateNew is the state returned from spdk_tgt. There is no underscore in the string.
+	SPDKShallowCopyStateInProgress = "in progress"
```
derekbit, line 76: It is annoying..

derekbit (approved):
LGTM

derekbit:
@shuo-wu It is a DRAFT. Will you convert it to READY?

derekbit (dismissed):

derekbit:
@shuo-wu will update the unit test later.

derekbit (approved):

PR #187, opened 2024-07-26
 feat(disk driver): don't explicitly specify userspace I/O driver while creating disk

derekbit:

Issue longhorn/longhorn#8813

- feat(disk driver): don't explicitly specify userspace I/O driver while creating disk
- fix(disk driver): fix deletion of disk based on nvme bdev driver


DamiaSan:
Is `setup.sh` that use `vfio-pci` as default?

derekbit:
> Is setup.sh that use vfio-pci as default?

Yes.

derekbit:
> Is `setup.sh` that use `vfio-pci` as default?

https://github.com/longhorn/spdk/blob/longhorn-v24.05/scripts/setup.sh#L412-L441

DamiaSan (approved):

PR #191, opened 2024-07-30
fix(disk): unable to unbind disk if it is failed to create it

derekbit:

Issue Longhorn/longhorn#9126

- If disk UUID is not provided while deleting a disk. Use disk name instead.
- Unbind a disk if failed to attach it while attaching.

### DIFF
--- a/pkg/spdk/disk.go
+++ b/pkg/spdk/disk.go
@@ -85,17 +85,26 @@ func svcDiskDelete(spdkClient *spdkclient.Client, diskName, diskUUID, diskPath, 
 		}
 	}()
 
-	if diskName == "" || diskUUID == "" {
-		return &emptypb.Empty{}, grpcstatus.Error(grpccodes.InvalidArgument, "disk name and disk UUID are required")
+	if diskName == "" {
+		return &emptypb.Empty{}, grpcstatus.Error(grpccodes.InvalidArgument, "disk name is required")
 	}
 
+	var lvstores []spdktypes.LvstoreInfo
 	bdevName := ""
-	lvstores, err := spdkClient.BdevLvolGetLvstore("", diskUUID)
+
+	if diskUUID != "" {
+		lvstores, err = spdkClient.BdevLvolGetLvstore("", diskUUID)
+	} else {
+		// The disk is not successfully created in creation stage because the diskUUID is not provided,
+		// so we blindly use the diskName as the bdevName here.
+		log.Warn("Disk UUID is not provided, trying to get lvstore with disk name")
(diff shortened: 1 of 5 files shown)


pkg/spdk/disk.go
```diff
@@ -85,25 +85,31 @@ func svcDiskDelete(spdkClient *spdkclient.Client, diskName, diskUUID, diskPath,
 		}
 	}()
 
-	if diskName == "" || diskUUID == "" {
-		return &emptypb.Empty{}, grpcstatus.Error(grpccodes.InvalidArgument, "disk name and disk UUID are required")
+	if diskName == "" {
```
shuo-wu, line 88: What if somehow only `diskUUID` is provided? 
The function should try its best to do the cleanup:
1. If `diskName` is provided, we will use this name as lvstore name during deletion
2. If `diskUUID` is provided, we will use this UUID as lvstoreUUID during deletion 
3. If both are provided, we will verify there is an exact lvstore matching both during deletion

derekbit: > What if somehow only diskUUID is provided?

Only `diskUUID` should be impossible, because the diskName is must-have information, so we only need to handle empty UUID case.
Ref: https://github.com/longhorn/longhorn-manager/blob/master/controller/node_controller.go#L1680-L1698

pkg/spdk/disk.go
```diff
@@ -85,17 +85,26 @@ func svcDiskDelete(spdkClient *spdkclient.Client, diskName, diskUUID, diskPath,
...
+		// so we blindly use the diskName as the bdevName here.
+		log.Warn("Disk UUID is not provided, trying to get lvstore with disk name")
+		lvstores, err = spdkClient.BdevLvolGetLvstore(diskName, "")
+	}
 	if err != nil {
 		if !jsonrpc.IsJSONRPCRespErrorNoSuchDevice(err) {
 			return nil, errors.Wrapf(err, "failed to get lvstore with UUID %v", diskUUID)
 		}
-		log.Warnf("Cannot find lvstore with UUID %v", diskUUID)
+		log.WithError(err).Warnf("Failed to get lvstore with UUID or disk name, blindly use disk name as bdev name")
```
c3y1huang, line 107:
```suggestion
		log.WithError(err).Warn("Failed to get lvstore with UUID or disk name, blindly use disk name as bdev name")
```

c3y1huang (changes requested):

pkg/spdk/disk.go
derekbit: Updated. Thank you.

c3y1huang (approved):

PR #25, opened 2023-05-16
Update Engine

shuo-wu:
1. Introduce engine cache
2. Implement API `EngineList`
3. Add monitoring for engine
4. Improve engine CURD implementation

PR #192, opened 2024-07-31
feat(v2): DR volume

c3y1huang:

Issue longhorn/longhorn#6613

Implement incremental restore.

`None`

`None`


go.mod
```diff
@@ -4,6 +4,8 @@ go 1.22.0
 
 toolchain go1.22.5
 
+replace github.com/longhorn/backupstore => github.com/c3y1huang/backupstore v0.0.0-20240731031751-cb5deaa2ca0e
```
c3y1huang, line 7: TODO: update after https://github.com/longhorn/backupstore/pull/183

pkg/spdk/replica.go
```diff
@@ -1925,14 +1925,18 @@ func (r *Replica) BackupRestore(spdkClient *spdkclient.Client, backupUrl, snapsh
...
+		r.log.Info("Resetting the restore for backup %v", backupUrl)
+
 		var lvolName string
 		var snapshotNameToBeRestored string
 
 		validLastRestoredBackup := r.canDoIncrementalRestore(restore, backupUrl, backupName)
 		if validLastRestoredBackup {
+			r.log.Infof("Starting an incremental restore for backup %v", backupUrl)
 			lvolName = GetReplicaSnapshotLvolName(r.Name, restore.LastRestored)
 			snapshotNameToBeRestored = restore.LastRestored
```
shuo-wu, line 1937: Do we need to change `restore.SnapshotName` and `restore.LvolName` during incremental restore? And isn't `restore.LastRestored` a **backup** name? IIUC, DR volume will store all data into a single **snapshot** lvol.

shuo-wu:
I am thinking if we can improve the restore flow for SPDK volumes.

In the beginning, unlike v1 volumes, we ask the engine to get involved during restoration. But later on, the engine still asks each replica to fetch data from the backups separately, which is quite inefficient. (While v1 volumes have to do replica-level restoration since their engines know little about backup & restore.) Would we directly start the whole engine with all available replicas? Then the data can be directly written into the frontend

DamiaSan:
> But later on, the engine still asks each replica to fetch data from the backups separately

@shuo-wu  Could you describe what you mean by _separately_?

pkg/spdk/restore.go
```diff
@@ -141,7 +141,7 @@ func (r *Restore) OpenVolumeDev(volDevName string) (*os.File, string, error) {
 	if err != nil {
 		return nil, "", errors.Wrapf(err, "failed to create NVMe initiator for lvol bdev %v", lvolName)
 	}
-	if _, err := initiator.Start(r.ip, strconv.Itoa(int(r.port)), false); err != nil {
+	if _, err := initiator.Start(r.ip, strconv.Itoa(int(r.port)), dmDeviceAndEndpointCleanupRequired); err != nil {
```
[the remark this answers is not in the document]
c3y1huang: After discussion with @derekbit @shuo-wu , we can blindly set `dmDeviceAndEndpointCleanupRequired` to `true` for both full restore and incremental restore.

pkg/spdk/replica.go
c3y1huang: After discussion with @derekbit @shuo-wu , a single snapshot needs to be taken and retained for incremental restore. I will check the snapshot naming during implementation.

pkg/spdk/replica.go
```diff
@@ -2050,6 +2083,36 @@ func (r *Replica) waitForRestoreComplete() error {
...
+
+	if r.restore.State == btypes.ProgressStateCanceled {
+		r.log.Info("Doing nothing for canceled backup restoration")
+		return nil
+	}
+
+	r.log.Infof("Deleting snapshot %v for snapshot replacement of the restored volume", restore.SnapshotName)
+	_, err := r.SnapshotDelete(spdkClient, restore.SnapshotName)
+	if err != nil {
+		r.log.WithError(err).Error("Failed to take snapshot of the restored volume")
```
DamiaSan, line 2097: Here you are deleting a snapshot

pkg/spdk/engine.go
```diff
@@ -1797,6 +1805,62 @@ func (e *Engine) BackupRestore(spdkClient *spdkclient.Client, backupUrl, engineN
...
+}
+
+func (e *Engine) waitForRestoreComplete() error {
+	periodicChecker := time.NewTicker(time.Duration(restorePeriodicRefreshInterval.Seconds()) * time.Second)
+	defer periodicChecker.Stop()
+
+	var err error
+	for range periodicChecker.C {
+		isReplicaRestoreCompleted := true
+		for replicaName, replicaAddress := range e.ReplicaAddressMap {
```
shuo-wu, line 1826: It's better to verify if the replica is mode RW before checking the restore status

pkg/spdk/engine.go
```diff
@@ -1797,6 +1805,62 @@ func (e *Engine) BackupRestore(spdkClient *spdkclient.Client, backupUrl, engineN
 	return resp, nil
 }
 
+func (e *Engine) completeBackupRestore(spdkClient *spdkclient.Client) error {
+	if err := e.waitForRestoreComplete(); err != nil {
+		return errors.Wrapf(err, "failed to wait for restore complete")
+	}
+
+	return e.BackupRestoreFinish(spdkClient)
```
shuo-wu, line 1816: If there is a goroutine wait here then automatically invoke `e.BackupRestoreFinish`, do we still need to expose gRPC API `EngineBackupRestoreFinish`?

Besides, I just notice that `e.BackupRestoreFinish` will attach all replicas and create a RAID (but without any further operations like checksum verification). What is the purpose? In other words, do we really need `e.BackupRestoreFinish`?

pkg/spdk/engine.go
```diff
@@ -61,7 +61,8 @@ type Engine struct {
 	Head        *api.Lvol
 	SnapshotMap map[string]*api.Lvol
 
-	IsRestoring bool
+	IsRestoring           bool
+	RestoringSnapshotName string
```
shuo-wu, line 65: For a DR volume, is this the final snapshot name, or the snapshot name of the upcoming incremental restore?

pkg/spdk/replica.go
```diff
@@ -2083,7 +2083,32 @@ func (r *Replica) waitForRestoreComplete() error {
...
+func (r *Replica) postIncrementalRestoreOperations(spdkClient *spdkclient.Client, restore *Restore) error {
+	r.log.Infof("Replacing snapshot %v of the restored volume", restore.SnapshotName)
+
+	if r.restore.State == btypes.ProgressStateCanceled {
+		r.log.Info("Doing nothing for canceled backup restoration")
+		return nil
+	}
+
+	r.log.Infof("Deleting snapshot %v for snapshot replacement of the restored volume", restore.SnapshotName)
+	_, err := r.SnapshotDelete(spdkClient, restore.SnapshotName)
```
shuo-wu, line 2095: What if this is the first snapshot of a DR volume?

shuo-wu (commented):
As we discussed in the meeting, do we need an API for DR volume activation?

pkg/spdk/engine.go
c3y1huang: Final and upcoming, only this snapshot name will be used during the restore.

pkg/spdk/replica.go
c3y1huang: AFAIK if we delete the first snapshot it will get coalesced into the head. Any concerns?

pkg/spdk/engine.go
c3y1huang: > If there is a goroutine wait here then automatically invoke `e.BackupRestoreFinish`, do we still need to expose gRPC API `EngineBackupRestoreFinish`?
> 

After discussing with @derekbit , we agreed that we don't need to expose gRPC API `EngineBackupRestoreFinish`, as it's not currently used by either v1 or v2 engine. Originally, this API was to abstract the finish operation.

> Besides, I just notice that `e.BackupRestoreFinish` will attach all replicas and create a RAID (but without any further operations like checksum verification). What is the purpose? In other words, do we really need `e.BackupRestoreFinish`?

At the start of the backup restore operation, we deleted the RAID bdev, disconnected all replicas, and set the e.IsRestoring flag. The purpose of `e.BackupRestoreFinish` is to complete the cycle by reattaching the replicas, recreating the RAID bdev, and resetting the flag.

c3y1huang: [Chore: remove EngineBackupRestoreFinish RPC #25](https://github.com/longhorn/types/pull/25).

pkg/spdk/replica.go
shuo-wu: Oh YES. SPDK can coalesce snapshot content into the current head lvol. Plz add a comment for this.

pkg/spdk/engine.go
shuo-wu: I see. Although launching a RAID for a restoring/DR volume is redundant, we don't need to modify the periodic verification logic.

pkg/spdk/replica.go
c3y1huang: Added comment, thank you.

pkg/spdk/engine.go
shuo-wu: Does the engine need to cache this? At least for now, it is used in `Engine. BackupRestore` only

c3y1huang: We will include e.RestoringSnapshotName in the BackupRestoreRequest to ensure the snapshot name is consistent.

shuo-wu: OK. It make sure DR volume snapshot name keeps unchanged

shuo-wu (approved):

DamiaSan (approved):

shuo-wu (approved):

pkg/spdk/engine.go
```diff
@@ -1748,15 +1749,28 @@ func (e *Engine) BackupRestore(spdkClient *spdkclient.Client, backupUrl, engineN
 	e.IsRestoring = true
 
 	// TODO: support DR volume
```
derekbit, line 1751: Remove `// TODO: support DR volume`

derekbit (commented):
LGTM.
Only one minor comment.

pkg/spdk/engine.go
c3y1huang: Fixed, thank you.

derekbit (approved):
Thanks @c3y1huang. Awesome task.

PR #194, opened 2024-08-03
fix: replicas should do expose blindly during rebuilding; server should not create orphan replicas for rebuilding lvols

shuo-wu:

Issue 
longhorn/longhorn#7199
longhorn/longhorn#9121

### DIFF
--- a/pkg/spdk/replica.go
+++ b/pkg/spdk/replica.go
@@ -111,7 +111,6 @@ type RebuildingSrcCache struct {
 	dstReplicaName string
 	// dstRebuildingBdev is the result of attaching the rebuilding lvol exposed by the dst replica
 	dstRebuildingBdevName string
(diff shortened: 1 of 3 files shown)


derekbit:
The "Invalid argument" issue stills exists.
```
FAIL: spdk_test.go:397: TestSuite.TestSPDKMultipleThreadSnapshotOpsAndRebuilding

spdk_test.go:1208:
    c.Assert(rebuildingStatus.Error, Equals, "")
... obtained string = "Invalid argument"
... expected string = ""
```

shuo-wu:
Weird. It's hard to trigger any errors in my private env, especially running test `TestSPDKMultipleThreadSnapshotOpsAndRebuilding` only

derekbit (approved):
LGTM

PR #195, opened 2024-08-08
fix(engine): mark replica as ERROR if it is not found

derekbit:

Issue https://github.com/longhorn/longhorn/issues/9166

Mark replica as ERROR if it is not found. Then, longhorn-manager will be aware of the failed replica and trigger cleanup.


pkg/spdk/engine.go
```diff
@@ -648,7 +648,12 @@ func (e *Engine) checkAndUpdateInfoFromReplicaNoLock() {
 		}
 		replica, err := replicaServiceCli.ReplicaGet(replicaName)
 		if err != nil {
-			e.log.WithError(err).Warnf("failed to get replica %s with address %s, will skip this replica and continue info update from replica", replicaName, address)
+			if strings.Contains(err.Error(), "cannot find replica") {
+				e.log.WithError(err).Warnf("Failed to get replica %s with address %s, make it as ERR", replicaName, address)
+				e.ReplicaModeMap[replicaName] = types.ModeERR
+			} else {
+				e.log.WithError(err).Warnf("failed to get replica %s with address %s, will skip this replica and continue info update from replica", replicaName, address)
```
shuo-wu, line 655: NIT: Correct the printed error log format then it's good to go

shuo-wu (dismissed):

pkg/spdk/engine.go
derekbit: Fixed. Thank you.

derekbit:
@shuo-wu There are some changes. Could you review it again? Thank you.

pkg/spdk/engine.go
```diff
@@ -648,7 +648,8 @@ func (e *Engine) checkAndUpdateInfoFromReplicaNoLock() {
 		}
 		replica, err := replicaServiceCli.ReplicaGet(replicaName)
 		if err != nil {
-			e.log.WithError(err).Warnf("failed to get replica %s with address %s, will skip this replica and continue info update from replica", replicaName, address)
+			e.log.WithError(err).Warnf("Failed to get replica %s with address %s, make it as ERR", replicaName, address)
```
shuo-wu, line 651: NIT: `mark the mode from %v to ERR`

This would provide a little bit more info

shuo-wu (approved):

innobead:
Well done. LGTM.

PR #199, opened 2024-08-10
fix: fix engine replica deletion and replica varification issues

shuo-wu:

### DIFF
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -643,12 +643,17 @@ func (e *Engine) checkAndUpdateInfoFromReplicaNoLock() {
 	// hasBackingImage := false
 	hasSnapshot := false
 	for replicaName, address := range e.ReplicaAddressMap {
-		if e.ReplicaModeMap[replicaName] != types.ModeRW {
+		if e.ReplicaModeMap[replicaName] != types.ModeRW && e.ReplicaModeMap[replicaName] != types.ModeWO {
+			if e.ReplicaModeMap[replicaName] != types.ModeERR {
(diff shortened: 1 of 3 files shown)


pkg/spdk/engine.go
```diff
@@ -1234,6 +1234,19 @@ func (e *Engine) ReplicaDelete(spdkClient *spdkclient.Client, replicaName, repli
...
+		if err = e.initiator.Suspend(true, true); err != nil {
+			return errors.Wrapf(err, "failed to suspend NVMe initiator before removing replica %s base bdev from engine %s RAID", replicaName, e.Name)
+		}
+		defer func() {
+			if err = e.initiator.Resume(); err != nil {
+				err = errors.Wrapf(err, "failed to suspend NVMe initiator after removing replica %s base bdev from engine %s RAID", replicaName, e.Name)
+			}
+		}()
+	}
+
```
derekbit, line 1252: What do we need to suspend frontend for the replica deletion?

derekbit:
@shuo-wu Can the PR be reviewed?

pkg/spdk/engine.go
shuo-wu: Let's remove this commit. 
When Longhorn deletes a replica from the engine, spdk_tgt will print out lots of errors like `nvme_tcp.c:2177:nvme_tcp_qpair_process_completions: *ERROR*: Failed to flush tqpair=0x1826e70 (9): Bad file descriptor`. I was thinking adding a frontend suspend would avoid it. but after testing it today, the answer is NO.

derekbit (approved):

derekbit (approved):
LGTM

derekbit (approved):

PR #200, opened 2024-08-11
fix: port leakage caused by spdk client

derekbit:

Issue Longhorn/longhorn#9190


innobead (approved):
LGTM.

I assume the client is a local variable in stack, so the connection created by it should be eventually garbage collected. However, the time should be uncertain, so the issue we encountered might happen in stress testing.

innobead:
Just noticed the arm64 build failed. Need to check if the tests are unstable.

derekbit:
> Just noticed the arm64 build failed. Need to check if the tests are unstable.

Yes, the tests are not stable. Checking it.

PR #203, opened 2024-08-12
refactor: refactor the codes for closing spdk clients

derekbit:

Issue longhorn/longhorn#9190

- fix(test): temporarily create spdk_tgt for each test case
- refactor: refactor the codes for closing spdk clients
    - Refactor the codes for closing spdk clients
    - Log error returned by the close function

### DIFF
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -215,11 +215,16 @@ func (e *Engine) Create(spdkClient *spdkclient.Client, replicaAddressMap map[str
 		e.IP = targetIP
 
 		// Get ReplicaModeMap and ReplicaBdevNameMap
-		targetSPDKClient, err := GetServiceClient(net.JoinHostPort(e.IP, strconv.Itoa(types.SPDKServicePort)))
+		targetSPDKServiceAddress := net.JoinHostPort(e.IP, strconv.Itoa(types.SPDKServicePort))
+		targetSPDKClient, err := GetServiceClient(targetSPDKServiceAddress)
 		if err != nil {
 			return nil, err
 		}
-		defer targetSPDKClient.Close()
+		defer func() {
+			if errClose := targetSPDKClient.Close(); errClose != nil {
+				e.log.WithError(errClose).Errorf("Failed to close target spdk client with address %s during create engine", targetSPDKServiceAddress)
+			}
+		}()
 
 		var engineWithTarget *api.Engine
 		if initiatorIP != targetIP {
@@ -642,6 +647,7 @@ func (e *Engine) checkAndUpdateInfoFromReplicaNoLock() {
 	replicaAncestorMap := map[string]*api.Lvol{}
 	// hasBackingImage := false
 	hasSnapshot := false
+
 	for replicaName, address := range e.ReplicaAddressMap {
 		if e.ReplicaModeMap[replicaName] != types.ModeRW && e.ReplicaModeMap[replicaName] != types.ModeWO {
 			if e.ReplicaModeMap[replicaName] != types.ModeERR {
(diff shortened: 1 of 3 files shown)


shuo-wu:
This race fix is already in https://github.com/longhorn/longhorn-spdk-engine/pull/199

pkg/spdk/server.go
```diff
@@ -251,19 +251,19 @@ func (s *Server) verify() (err error) {
 		func() {
 			r.RLock()
 			defer r.RUnlock()
-		if r.ChainLength < 2 {
-			delete(replicaMap, replicaName)
+			if r.ChainLength < 2 {
```
shuo-wu, line 254: Remove the below changes from this commit

pkg/spdk/engine.go
```diff
@@ -651,58 +657,66 @@ func (e *Engine) checkAndUpdateInfoFromReplicaNoLock() {
 			e.log.WithError(err).Warnf("failed to get service client for replica %s with address %s, will skip this replica and continue info update from replica", replicaName, address)
```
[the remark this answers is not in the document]
derekbit: Yes, we can, but there is no difference. WDYT?

shuo-wu:
Are there any reasons we decide to revert to the previous spdk_tgt launch for the unit test?

pkg/spdk/engine.go
```diff
@@ -642,6 +647,7 @@ func (e *Engine) checkAndUpdateInfoFromReplicaNoLock() {
 	replicaAncestorMap := map[string]*api.Lvol{}
 	// hasBackingImage := false
 	hasSnapshot := false
+
 	for replicaName, address := range e.ReplicaAddressMap {
 		if e.ReplicaModeMap[replicaName] != types.ModeRW {
```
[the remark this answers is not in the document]
derekbit: Can you elaborate more on it? The purpose of the change is to simplify `close` of spdk client.

pkg/spdk/engine.go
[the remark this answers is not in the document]
shuo-wu: I prefer to move the client open & close into the same logic block. Anyway, it indeed functions the same.

pkg/spdk/engine.go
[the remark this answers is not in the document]
shuo-wu: This makes the below WO replica checks unreachable. It's my bad. The previous is buggy
```
if e.ReplicaModeMap[replicaName] != types.ModeRW {
	continue
}
......
if e.ReplicaModeMap[replicaName] == types.ModeWO {
	......
}
```

derekbit:
> Are there any reasons we decide to revert to the previous spdk_tgt launch for the unit test?

Just a temporary change for ensuring each test case works expectedly.
The successive test cases seem unstable and sometime fail. We can check the failure later.

pkg/spdk/server.go
derekbit: Use your change instead.

pkg/spdk/engine.go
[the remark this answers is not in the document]
derekbit: No prob!

pkg/spdk/engine.go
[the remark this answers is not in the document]
derekbit: Rebased. Thank you.

pkg/spdk/engine.go
```diff
@@ -215,11 +215,16 @@ func (e *Engine) Create(spdkClient *spdkclient.Client, replicaAddressMap map[str
...
-		targetSPDKClient, err := GetServiceClient(net.JoinHostPort(e.IP, strconv.Itoa(types.SPDKServicePort)))
+		targetSPDKServiceAddress := net.JoinHostPort(e.IP, strconv.Itoa(types.SPDKServicePort))
+		targetSPDKClient, err := GetServiceClient(targetSPDKServiceAddress)
 		if err != nil {
 			return nil, err
 		}
-		defer targetSPDKClient.Close()
+		defer func() {
+			if errClose := targetSPDKClient.Close(); errClose != nil {
+				e.log.WithError(errClose).Errorf("Failed to close target spdk client with address %s", targetSPDKServiceAddress)
```
shuo-wu, line 225: NIT: `"Failed to .... during engine create"`
Similar below.

This helps quickly pinpoint which function has trouble with closing replica clients

shuo-wu (approved):
After improve the log, this PR is good to go

pkg/spdk/engine.go

derekbit:
Ready for review cc @shuo-wu  @innobead

pkg/spdk/engine.go
```diff
@@ -1299,10 +1299,14 @@ func (e *Engine) ReplicaDelete(spdkClient *spdkclient.Client, replicaName, repli
...
 
 	// Detaching the corresponding NVMf controller to remote replica
-	controllerName := helperutil.GetNvmeControllerNameFromNamespaceName(e.ReplicaBdevNameMap[replicaName])
-	e.log.Infof("Detaching the corresponding NVMf controller %v during remote replica %s delete", controllerName, replicaName)
-	if _, err := spdkClient.BdevNvmeDetachController(controllerName); err != nil && !jsonrpc.IsJSONRPCRespErrorNoSuchDevice(err) {
-		return errors.Wrapf(err, "failed to detach controller %s for deleting replica %s", controllerName, replicaName)
+	if _, ok := e.ReplicaBdevNameMap[replicaName]; ok {
+		controllerName := helperutil.GetNvmeControllerNameFromNamespaceName(e.ReplicaBdevNameMap[replicaName])
+		if controllerName != "" {
+			e.log.Infof("Detaching the corresponding NVMf controller %v during remote replica %s delete", controllerName, replicaName)
```
shuo-wu, line 1305: I was thinking blindly detaching the controller for the replica was better, no matter if there is a record in `e.ReplicaBdevNameMap`. Any reason why we need to add this check?

derekbit: If `controllerName` is empty string, there are lots of annoying error messages in log.

derekbit: @shuo-wu Any concerns?

shuo-wu: I don't think this will be often triggered. I am more worried about the case that the controller is there but somehow `e.ReplicaBdevNameMap[replicaName]` is empty

derekbit: I see. Removed the commit. Thank you.

shuo-wu (approved):

PR #205, opened 2024-08-14
fix: no need to empty bdev name in ReplicaBdevNameMap

derekbit:

Issue Longhorn/longhorn#9190

If the bdev name in ReplicaBdevNameMap is missing, BdevRaidRemoveBaseBdev has no chances to delete the base bdev.

### DIFF
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -191,7 +191,7 @@ func (e *Engine) Create(spdkClient *spdkclient.Client, replicaAddressMap map[str
 			if err != nil {
 				e.log.WithError(err).Warnf("Failed to get bdev from replica %s with address %s during creation, will mark the mode from %v to ERR and skip it and continue", replicaName, replicaAddr, e.ReplicaModeMap[replicaName])
 				e.ReplicaModeMap[replicaName] = types.ModeERR
-				e.ReplicaBdevNameMap[replicaName] = ""
+				e.ReplicaBdevNameMap[replicaName] = bdevName
 				continue
 			}
 			// TODO: Check if a replica is really a RW replica rather than a rebuilding failed replica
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM

pkg/spdk/engine.go
```diff
@@ -191,7 +191,7 @@ func (e *Engine) Create(spdkClient *spdkclient.Client, replicaAddressMap map[str
 			if err != nil {
 				e.log.WithError(err).Warnf("Failed to get bdev from replica %s with address %s during creation, will mark the mode from %v to ERR and skip it and continue", replicaName, replicaAddr, e.ReplicaModeMap[replicaName])
 				e.ReplicaModeMap[replicaName] = types.ModeERR
-				e.ReplicaBdevNameMap[replicaName] = ""
+				e.ReplicaBdevNameMap[replicaName] = bdevName
```
shuo-wu, line 194: This one seems meaningless. `connectNVMfBdev` failure would return an empty bdev name

pkg/spdk/engine.go
```diff
@@ -570,7 +569,6 @@ func (e *Engine) ValidateAndUpdate(spdkClient *spdkclient.Client) (err error) {
 	// Verify engine map consistency
 	for replicaName := range e.ReplicaAddressMap {
 		if _, exists := e.ReplicaBdevNameMap[replicaName]; !exists {
-			e.ReplicaBdevNameMap[replicaName] = ""
```
shuo-wu, line 573: I am not sure if this modification is OK since I want to make sure all maps hold the same key set after the validation.

derekbit:
After discussing with @shuo-wu, we will improve the address, mod and bdev maps in the development of v1.8.0, but we will have some improvements in https://github.com/longhorn/longhorn-spdk-engine/pull/206.

cc @innobead

PR #208, opened 2024-08-17
refactor: engine consolidates the replica related maps into a single one

shuo-wu:

Issue longhorn/longhorn#9244

### DIFF
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -47,10 +47,7 @@ type Engine struct {
 	Nqn        string
 	Nguid      string
 
-	// TODO: Use a single map to store all replica info
-	ReplicaAddressMap  map[string]string
-	ReplicaBdevNameMap map[string]string
-	ReplicaModeMap     map[string]types.Mode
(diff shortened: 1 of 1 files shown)


pkg/spdk/engine.go
```diff
@@ -1442,11 +1415,14 @@ func (e *Engine) snapshotOperation(spdkClient *spdkclient.Client, inputSnapshotN
 
 func (e *Engine) getReplicaClients() (replicaClients map[string]*client.SPDKClient, err error) {
 	replicaClients = map[string]*client.SPDKClient{}
-	for replicaName := range e.ReplicaAddressMap {
-		if e.ReplicaModeMap[replicaName] == types.ModeERR {
+	for replicaName, replicaStatus := range e.ReplicaStatusMap {
+		if replicaStatus.Mode != types.ModeRW && replicaStatus.Mode != types.ModeWO {
```
james-munson, line 1419: Slight logic change here in the case that Mode is unset and is "".  But perhaps better, because it will be skipped, just as for ModeERR.

derekbit (approved):
LGTM

PR #241, opened 2024-11-17
feat(v2 upgrade): support engine live upgrade

derekbit:

Issue Longhorn/longhorn#9104

Signed-off-by: Derek Su <derek.su@suse.com>


DamiaSan (approved):
LGTM

PR #251, opened 2024-12-01
feat: decrease transport_ack_timeout

derekbit:

Issue Longhorn/longhorn#9874

Signed-off-by: Derek Su <derek.su@suse.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/longhorn/backupstore v0.0.0-20241130163459-2b482603a2c6
 	github.com/longhorn/go-common-libs v0.0.0-20241128023039-4d6c3a880dbc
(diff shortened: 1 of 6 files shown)


DamiaSan:
Just one question: shouldn't we modify also the constants defined in `go-spdk-helper\pkg\nvme\nvmecli.go`?

DamiaSan:
> > Just one question: shouldn't we modify also the constants defined in `go-spdk-helper\pkg\nvme\nvmecli.go`?
> 
> Which const?

```
	defaultCtrlLossTmo    = 30
	defaultKeepAliveTmo   = 5
	defaultReconnectDelay = 2
```
These are used by the nvme cli initiator, shouldn't it be affected by the same issue?

derekbit:
> These are used by the nvme cli initiator, shouldn't it be affected by the same issue?

No, the change of transport_ack_timeout is for the SPDK's transport layer and is nothing to do with the three parameters you mentioned.

DamiaSan (approved):

PR #263, opened 2024-12-10
fix: replica prepares a head with the caches correctly

PhanLe1010:
This fixes a corner case: A replica has a correct head lvol but its caches get missed.

Now r.prepareHead() will create a head lvol if it is invalid or does not exists, then blindly update the caches.

Issue # https://github.com/longhorn/longhorn/issues/9942


shuo-wu (approved):

derekbit (approved):
LGTM

PR #278, opened 2024-12-20
feat: v2 fast rebuilding

shuo-wu:
#### Which issue(s) this PR fixes:

https://github.com/longhorn/longhorn/issues/9488

The related docs that can help understand this feature:
[V2 Data Engine Fast Rebuilding diagram flow](https://docs.google.com/drawings/d/1670SGum9uxZPJvw4Rd5WE4-7FssYCMPljmoix-3sif8/edit?usp=sharing)
[V2 Data Engine Fast Rebuilding illustration and animation](https://docs.google.com/presentation/d/15AHms4wGytgYu4_yJoKdfYHHV6ZqAbWFV-ZPLAmMpMM/edit?usp=sharing)

PR #284, opened 2024-12-25
fix: reconstruct ActiveChain and SnapshotLvolMap while creating replica

derekbit:

Issue Longhorn/longhorn#10033

If the replica's ActiveChain is not constructed while creating the replica, the mismatch found by the `validateAndUpdate()` will result in the error state.
    
    ```
    ....
            newSnapshotLvolMap, err := constructSnapshotLvolMap(r.Name, bdevLvolMap)
            if err != nil {
                    return err
            }
            if len(r.SnapshotLvolMap) != len(newSnapshotLvolMap) {
                    return fmt.Errorf("replica current active snapshot ....")
            }
    ...
    ````

### DIFF
--- a/pkg/spdk/replica.go
+++ b/pkg/spdk/replica.go
@@ -249,7 +249,7 @@ func (r *Replica) construct(bdevLvolMap map[string]*spdktypes.BdevInfo) (err err
 	}()
 
 	switch r.State {
-	case types.InstanceStatePending:
+	case types.InstanceStateStopped, types.InstanceStatePending:
 		break
 	case types.InstanceStateRunning:
 		if r.isRebuilding {
@@ -735,6 +735,16 @@ func (r *Replica) Create(spdkClient *spdkclient.Client, portCount int32, superio
 	if err := r.prepareHead(spdkClient, backingImage); err != nil {
 		return nil, err
 	}
+
+	// Construct the snapshot lvol map and the active chain
+	bdevLvolMap, err := GetBdevLvolMapWithFilter(spdkClient, r.replicaLvolFilter)
+	if err != nil {
+		return nil, err
+	}
+	if err := r.construct(bdevLvolMap); err != nil {
+		return nil, err
+	}
+
 	r.State = types.InstanceStateStopped
 
 	podIP, err := commonnet.GetIPForPod()
--- a/pkg/spdk/server.go
+++ b/pkg/spdk/server.go
@@ -3,7 +3,6 @@ package spdk
 import (
 	"fmt"
 	"net"
-	"reflect"
 	"strconv"
 	"sync"
 	"time"
@@ -332,7 +331,7 @@ func (s *Server) verify() (err error) {
 			}
 		}
 	}
-	if !reflect.DeepEqual(s.replicaMap, replicaMap) {
+	if len(s.replicaMap) != len(replicaMap) {
 		logrus.Infof("spdk gRPC server: Replica map updated, map count is changed from %d to %d", len(s.replicaMap), len(replicaMap))
 	}
 	s.replicaMap = replicaMap


derekbit:
The failure will be fixed in https://github.com/longhorn/longhorn-spdk-engine/pull/278

derekbit:
After applying the fix, the [e2e test can successfully run 8 times](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2314/console) without hitting the errors, but it hits
```
Writing random data to pod e2e-test-deployment-0-6456f6c484-przqc failed with error (0)
Reason: Handshake status 500 Internal Server Error -+-+- {'content-length': '29', 'content-type': 'text/plain; charset=utf-8', 'date': 'Wed, 25 Dec 2024 14:38:58 GMT'} -+-+- b'container not found ("sleep")'
```

The error has nothing to do with the issue. It will be investigated in  https://github.com/longhorn/longhorn/issues/8207.

cc @shuo-wu @c3y1huang @yangchiu @innobead

innobead:
Because the construction happens when creating a replica, it becomes a blocking operation. Do you think this will be a concern? if multiple replicas are created and re-constructed at the same time.

innobead (approved):
LGTM.

We might need to consider whether the current implementation is scalable.

innobead:
> ```
> Writing random data to pod e2e-test-deployment-0-6456f6c484-przqc failed with error (0)
> Reason: Handshake status 500 Internal Server Error -+-+- {'content-length': '29', 'content-type': 'text/plain; charset=utf-8', 'date': 'Wed, 25 Dec 2024 14:38:58 GMT'} -+-+- b'container not found ("sleep")'
> ```

Do we know what server returned this error? Usually, it just means a server error.

derekbit:
> Because the construction happens when creating a replica, it becomes a blocking operation. Do you think this will be a concern? if multiple replicas are created and re-constructed at the same time.

It is per-replica lock, so multiple replicas shouldn't be an issue.


> Do we know what server returned this error? Usually, it just means a server error.

After investigating the support bundle, volume controller is unable to salvage because of the disk pressure ([codes](https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L1397-L1408)) due to the orphaned replicas created while restarting nodes.
The check is introduced by the https://github.com/longhorn/longhorn/issues/1974. This is a bit weird. Do we really need to give up the failed replica because of the disk pressure? cc @shuo-wu @PhanLe1010 @innobead

innobead:
> After investigating the support bundle, volume controller is unable to salvage because of the disk pressure ([codes](https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L1397-L1408)) due to the orphaned replicas created while restarting nodes. The check is introduced by the [longhorn/longhorn#1974](https://github.com/longhorn/longhorn/issues/1974). This is a bit weird. Do we really need to give up the failed replica because of the disk pressure? cc @shuo-wu @PhanLe1010 @innobead

This issue seems related to the current implementation, where failed replicas cannot be reused. The failed salvage due to unused orphaned replicas appears to be expected in this context. 

We should revisit this once we have complete snapshot reuse in version 1.9, including intact snapshot reuse (introduced in 1.8) and delta snapshot reuse (planned for 1.9). cc @longhorn/dev-data-plane

derekbit:
Added a support bundle for the future improvement
[supportbundle_f141eeb1-d42b-4def-a700-b189bbe84ac6_2024-12-25T13-34-39Z.zip](https://github.com/user-attachments/files/18249282/supportbundle_f141eeb1-d42b-4def-a700-b189bbe84ac6_2024-12-25T13-34-39Z.zip)

problematic volume: pvc-a9c26ee9-3918-4ef6-b243-c340b160b41b

yangchiu:
> After applying the fix, the [e2e test can successfully ran 8 times](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2314/console) without hitting the errors, but it hit
> 
> ```
> 
> Writing random data to pod e2e-test-deployment-0-6456f6c484-przqc failed with error (0)
> 
> Reason: Handshake status 500 Internal Server Error -+-+- {'content-length': '29', 'content-type': 'text/plain; charset=utf-8', 'date': 'Wed, 25 Dec 2024 14:38:58 GMT'} -+-+- b'container not found ("sleep")'
> 
> ```
> 
> 
> 
> The error has nothing to do with the issue. It will be investigated in  https://github.com/longhorn/longhorn/issues/8207.
> 
> 
> 
> cc @shuo-wu @c3y1huang @yangchiu @innobead 

@derekbit This message simply indicates that the test is waiting for a pod that isn't currently running. It doesn't draw any conclusions and requires human intervention to investigate any potential issues. Possible causes could include:

1. AWS instances entering an abnormal state following a reboot (infrastructure issue).
2. The pod taking an unexpectedly long time to be recreated after a node reboot, which can be resolved by increasing the timeout in the test code.
3. The pod getting stuck in the creating, pending, or terminating states, rather than running. There's a high chance that the corresponding volume isn't in an attached/healthy state (product issue).

Just for your information.

derekbit:
> > After applying the fix, the [e2e test can successfully ran 8 times](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2314/console) without hitting the errors, but it hit
> > ```
> > 
> > Writing random data to pod e2e-test-deployment-0-6456f6c484-przqc failed with error (0)
> > 
> > Reason: Handshake status 500 Internal Server Error -+-+- {'content-length': '29', 'content-type': 'text/plain; charset=utf-8', 'date': 'Wed, 25 Dec 2024 14:38:58 GMT'} -+-+- b'container not found ("sleep")'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The error has nothing to do with the issue. It will be investigated in  [longhorn/longhorn#8207](https://github.com/longhorn/longhorn/issues/8207).
> > cc @shuo-wu @c3y1huang @yangchiu @innobead
> 
> @derekbit This message simply indicates that the test is waiting for a pod that isn't currently running. It doesn't draw any conclusions and requires human intervention to investigate any potential issues. Possible causes could include:
> 
> 1. AWS instances entering an abnormal state following a reboot (infrastructure issue).
> 2. The pod taking an unexpectedly long time to be recreated after a node reboot, which can be resolved by increasing the timeout in the test code.
> 3. The pod getting stuck in the creating, pending, or terminating states, rather than running. There's a high chance that the corresponding volume isn't in an attached/healthy state (product issue).
> 
> Just for your information.

Thanks @yangchiu 
We have found the root cause. Please check https://github.com/longhorn/longhorn-spdk-engine/pull/284#issuecomment-2561951749.

PR #288, opened 2024-12-26
fix(server): fix race conditions

derekbit:

Issue Longhorn/longhorn#9953

- fix(server): add replica to replicaMap after replica creation and initialization
- fix race conditions in spdk_tgt.go
- fix(scripts): prevent different test suites from running concurrently
- fix(backingimage): increase HTTPTimeout to 10 seconds


ChanYiLin (approved):
LGTM

PR #292, opened 2024-12-28
revert "fix(scripts): prevent different test suites from running concurrently

shuo-wu:
This reverts commit 7081bdafab53a5d482b022d875993c80912aa632.

Somehow this fails unit tests in my dev machine. Need to do more investigation.


derekbit:
We will investigate if the parameter introduces any negative impact.

derekbit (approved):

PR #298, opened 2025-01-06
fix: replica address is missing in the backup status response

derekbit:

Issue Longhorn/longhorn#10136

Replica address is missing in the backup status response. The replica address is required for test_basic.py::test_backup_status_for_unavailable_replicas.

### DIFF
--- a/pkg/spdk/replica.go
+++ b/pkg/spdk/replica.go
@@ -209,6 +209,12 @@ func NewReplica(ctx context.Context, replicaName, lvsName, lvsUUID string, specS
 	}
 }
 
+func (r *Replica) GetAddress() string {
+	r.RLock()
+	defer r.RUnlock()
+	return net.JoinHostPort(r.IP, strconv.Itoa(int(r.PortStart)))
+}
+
 func (r *Replica) IsRebuilding() bool {
 	r.RLock()
 	defer r.RUnlock()
--- a/pkg/spdk/server.go
+++ b/pkg/spdk/server.go
@@ -1365,12 +1365,18 @@ func (s *Server) ReplicaBackupStatus(ctx context.Context, req *spdkrpc.BackupSta
 		return nil, grpcstatus.Errorf(grpccodes.NotFound, "cannot find backup %v", req.Backup)
 	}
 
+	replicaAddress := ""
+	if backup.replica != nil {
+		replicaAddress = fmt.Sprintf("tcp://%s", backup.replica.GetAddress())
+	}
+
 	return &spdkrpc.BackupStatusResponse{
-		Progress:     int32(backup.Progress),
-		BackupUrl:    backup.BackupURL,
-		Error:        backup.Error,
-		SnapshotName: backup.SnapshotName,
-		State:        string(backup.State),
+		Progress:       int32(backup.Progress),
+		BackupUrl:      backup.BackupURL,
+		Error:          backup.Error,
+		SnapshotName:   backup.SnapshotName,
+		State:          string(backup.State),
+		ReplicaAddress: replicaAddress,
 	}, nil
 }
 
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -1912,7 +1912,10 @@ func (e *Engine) BackupCreate(backupName, volumeName, engineName, snapshotName, 
 	e.log.Infof("Creating backup %s", backupName)
 
 	e.Lock()
-	defer e.Unlock()
+	defer func() {
+		e.Unlock()
+		e.UpdateCh <- nil
+	}()
 
 	replicaName, replicaAddress := "", ""
 	for name, replicaStatus := range e.ReplicaStatusMap {


pkg/spdk/server.go
```diff
@@ -1366,11 +1366,12 @@ func (s *Server) ReplicaBackupStatus(ctx context.Context, req *spdkrpc.BackupSta
...
-		BackupUrl:    backup.BackupURL,
-		Error:        backup.Error,
-		SnapshotName: backup.SnapshotName,
-		State:        string(backup.State),
+		Progress:       int32(backup.Progress),
+		BackupUrl:      backup.BackupURL,
+		Error:          backup.Error,
+		SnapshotName:   backup.SnapshotName,
+		State:          string(backup.State),
+		ReplicaAddress: req.ReplicaAddress,
```
innobead, line 1374: nit: This approach looks a little strange, as the response field is copied directly from the request. Ideally, the replica address should come from the backup info.

pkg/client/client.go
```diff
@@ -751,13 +751,14 @@ func (c *SPDKClient) EngineBackupStatus(backupName, engineName, replicaAddress s
...
-func (c *SPDKClient) ReplicaBackupStatus(backupName string) (*spdkrpc.BackupStatusResponse, error) {
+func (c *SPDKClient) ReplicaBackupStatus(backupName, replicaAddress string) (*spdkrpc.BackupStatusResponse, error) {
 	client := c.getSPDKServiceClient()
 	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceTimeout)
 	defer cancel()
 
 	return client.ReplicaBackupStatus(ctx, &spdkrpc.BackupStatusRequest{
-		Backup: backupName,
+		Backup:         backupName,
+		ReplicaAddress: replicaAddress,
```
innobead, line 761: Suggest using the replica address from the client directly instead of relying on another redundant parameter of replica address.

```go
	return client.ReplicaBackupStatus(ctx, &spdkrpc.BackupStatusRequest{
		Backup: backupName,
		ReplicaAddress: c.serviceURL,
	})
```

innobead (changes requested):

pkg/spdk/server.go
derekbit: Yes,it is a bit weird, but v1 uses the same method.
https://github.com/longhorn/longhorn-engine/blob/master/pkg/sync/backup.go#L133

innobead: Sounds good. Better to have an issue to track but low priority.

pkg/client/client.go
derekbit: Updated. Thank you.

pkg/spdk/server.go

pkg/client/client.go
```diff
@@ -751,7 +751,7 @@ func (c *SPDKClient) EngineBackupStatus(backupName, engineName, replicaAddress s
 	})
 }
 
-func (c *SPDKClient) ReplicaBackupStatus(backupName string) (*spdkrpc.BackupStatusResponse, error) {
+func (c *SPDKClient) ReplicaBackupStatus(backupName, replicaAddress string) (*spdkrpc.BackupStatusResponse, error) {
```
innobead, line 754: Do we still need this new parameter?

derekbit: Removed. Thanks.

innobead: Still exists?

derekbit: Forgot to push.

shuo-wu (dismissed):

pkg/spdk/server.go
```diff
@@ -1365,12 +1365,18 @@ func (s *Server) ReplicaBackupStatus(ctx context.Context, req *spdkrpc.BackupSta
 		return nil, grpcstatus.Errorf(grpccodes.NotFound, "cannot find backup %v", req.Backup)
 	}
 
+	replicaAddress := ""
+	if backup.replica != nil {
+		replicaAddress = fmt.Sprintf("tcp://%s:%d", backup.replica.IP, backup.replica.PortStart)
```
shuo-wu, line 1370: It's better to use a function like `replica.GetAddress()` rather than directly accessing the replica variables (without locking).

pkg/spdk/replica.go
```diff
@@ -209,6 +209,12 @@ func NewReplica(ctx context.Context, replicaName, lvsName, lvsUUID string, specS
 	}
 }
 
+func (r *Replica) GetAddress() string {
+	r.RLock()
+	defer r.RUnlock()
+	return fmt.Sprintf("tcp://%s:%d", r.IP, r.PortStart)
```
shuo-wu, line 215: NIT: To avoid confusion or misleading, I prefer to let the caller to add prefix `tcp://` , or change the function name

derekbit: Updated. Thank you.

shuo-wu (approved):

PR #301, opened 2025-01-07
chore(workflow): update runners to longhorn-infra-oracle-<arch>-runners

derekbit:

Issue Longhorn/longhorn#10145

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   build-amd64:
     name: Build AMD64 binaries
-    runs-on: longhorn-infra-amd64-runners
+    runs-on: longhorn-infra-oracle-amd64-runners
     steps:
     - name: Checkout code
       uses: actions/checkout@v4
@@ -28,7 +28,7 @@ jobs:
 
   build-arm64:
     name: Build ARM64 binaries
-    runs-on: longhorn-infra-arm64-runners
+    runs-on: longhorn-infra-oracle-arm64-runners
     steps:
     - name: Checkout code
       uses: actions/checkout@v4


.github/workflows/build.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   build-amd64:
     name: Build AMD64 binaries
-    runs-on: longhorn-infra-amd64-runners
+    runs-on: longhorn-infra-oracle-arm64-runners
```
innobead, line 10: Create a new one for amd64? Or just use a general runner from CNCF?

In the current infra, each runner is a pod/container, so 3 node infra cluster can run several runners. We might need to setup some similar on the new infra later.

innobead (approved):
How many runners we are running now?

.github/workflows/build.yml
derekbit: For SPDK, we need a customized runner for both arm64 and amd64 because of the necessary kernel modules and hugepages.

derekbit: > In the current infra, each runner is a pod/container, so 3 node infra cluster can run several runners. We might need to setup some similar on the new infra later.

Yes, we do the same thing in the oracle runner by https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/about-actions-runner-controller

derekbit:
> How many runners we are running now?

1 node for arm64 and 1 node for amd64

.github/workflows/build.yml
innobead: Did you install a single node cluster running this controller? on each runner host?

derekbit: single node cluster

innobead (approved):

PR #305, opened 2025-01-13
fix: skip write zero when dumping backing image data to lvol

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/9876

Since @derekbit has fixed the issue of spdk lvol init doesn't clean up the old data 
We can now skip writing the data when dumping backing image data to lvol


derekbit (approved):

derekbit:
LGTM

PR #311, opened 2025-01-24
feat: enable Open CAS  Framework

liyimeng:
Enable OCF https://github.com/Open-CAS, prepare to add caching to V2 data engine 

partially fix https://github.com/longhorn/longhorn/issues/10292

Issue #10292

Prepare for OCF support

This is a minor change, unlikely adding too much into the finally binary. So even we are not able to implement caching right now,  this won't add too much negative impact, I guess.


derekbit (approved):

derekbit:
Thanks @liyimeng. We will enable Open CAS in https://github.com/longhorn/dep-versions instead.

PR #314, opened 2025-02-05
fix: fix backing image size calculation

ChanYiLin:
ref: [longhorn/longhorn 10342](https://github.com/longhorn/longhorn/issues/10342)

Since we skip writing zero into the lvol when creating v2 backing image
the size of the backing image should be `bdevLvol.NumBlocks * uint64(bdevLvol.BlockSize)` instead of using `NumAllocatedClusters`

For example, parrot is 32 MB but it only allocates 6 clusters
We should report 32MB when getting the backing image size.

### DIFF
--- a/pkg/spdk/backing_image.go
+++ b/pkg/spdk/backing_image.go
@@ -131,7 +131,7 @@ func (bi *BackingImage) Create(spdkClient *spdkclient.Client, superiorPortAlloca
 		if err != nil {
 			bi.log.WithError(err).Error("Failed to create backing image")
 		}
-		// update the backing image afte preparing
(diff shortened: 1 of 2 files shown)


ChanYiLin:
Hi @shuo-wu 
Can you help to take a look on this PR, thanks!

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.8.x

shuo-wu (approved):

shuo-wu:
> @mergify backport v1.8.x

Do we need to create version branches for this repo?

PR #318, opened 2025-02-11
chore: versions of dependent packages are centrally managed in longhorn/dep-versions

derekbit:

Issue longhorn/longhorn#10208


derekbit:
Addressed in https://github.com/longhorn/longhorn-spdk-engine/pull/326

PR #319, opened 2025-02-14
fix: disable cksum calculation for system snapshots during and after rebuilding

shuo-wu:

Issue longhorn/longhorn#10335

so that the subsequent snapshot purge won't be blocked by checksum calculation.

### DIFF
--- a/pkg/spdk/replica.go
+++ b/pkg/spdk/replica.go
@@ -38,6 +38,8 @@ import (
 
 const (
 	restorePeriodicRefreshInterval = 2 * time.Second
+
+	checksumWaitPeriodAfterRebuilding = 10 * time.Second
 )
 
 type Replica struct {
@@ -78,6 +80,7 @@ type Replica struct {
 	// The rebuilding destination replica should cache this info
 	isRebuilding       bool
 	rebuildingDstCache RebuildingDstCache
+	lastRebuildingAt   time.Time
 
 	// The rebuilding source replica should cache this info
 	rebuildingSrcCache RebuildingSrcCache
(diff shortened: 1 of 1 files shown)


pkg/spdk/replica.go
```diff
@@ -249,6 +252,17 @@ func (r *Replica) Sync(spdkClient *spdkclient.Client) (err error) {
 			if bdevLvol.DriverSpecific.Lvol.Xattrs[spdkclient.SnapshotChecksum] != "" {
 				continue
 			}
+			parentBdevLvol := bdevLvolMap[bdevLvol.DriverSpecific.Lvol.BaseSnapshot]
+			if bdevLvol.DriverSpecific.Lvol.Xattrs[spdkclient.UserCreated] == "false" || (parentBdevLvol != nil && parentBdevLvol.DriverSpecific.Lvol.Xattrs[spdkclient.UserCreated] == "false") {
+				// Skip the checksum calculation of system created snapshot lvols during rebuilding as they may be purged later.
+				if r.isRebuilding {
```
DamiaSan, line 258: I am a little bit confused: do we skip or do we delay the checksum calculation of system created snapshot?

shuo-wu: At the end of each rebuilding, Longhorn will issue a snapshot purge request, which will try its best to remove all system snapshots. In other words, calculating these system snapshot checksums does not make sense as they will be removed soon. This delay would make the system snapshot deletion be handled before the calculation starts.

derekbit (approved):
LGTM

derekbit:
cc @c3y1huang

PR #326, opened 2025-03-06
chore(workflow): use oracle runners

derekbit:

Issue https://github.com/longhorn/longhorn/issues/10497

### DIFF
--- a/pkg/spdk_test.go
+++ b/pkg/spdk_test.go
@@ -65,6 +65,8 @@ var (
 	defaultTestSnapChecksumWaitCount    = 60
 
 	maxBackingImageGetRetries = 300
+
+	SPDKTGTBinary = "spdk_tgt"
 )
 
 func Test(t *testing.T) { TestingT(t) }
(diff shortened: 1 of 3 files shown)


derekbit:
CI failures will be tracked in other tickets
- https://github.com/longhorn/longhorn/issues/10598
- https://github.com/longhorn/longhorn/issues/10599
- ...

pkg/spdk_test.go
```diff
@@ -73,22 +75,19 @@ type TestSuite struct{}
...
-}
-
 func startTarget(spdkDir string, args []string, execute func(envs []string, binary string, args []string, timeout time.Duration) (string, error)) (err error) {
 	argsInStr := ""
 	for _, arg := range args {
 		argsInStr = fmt.Sprintf("%s %s", argsInStr, arg)
 	}
+
+	binary := SPDKTGTBinary
+	if spdkDir == "" {
```
innobead, line 85: Why do we need this condition?

I thought `filepath.Join` will handle it when `spdkDir` is empty.

pkg/spdk_test.go
```diff
@@ -105,7 +104,8 @@ func LaunchTestSPDKTarget(c *C, execute func(envs []string, name string, args []
 
 	if !targetReady {
 		go func() {
-			err := startTarget(GetSPDKDir(), []string{"--logflag all", "2>&1 | tee /tmp/spdk_tgt.log"}, execute)
+			logrus.Info("Starting SPDK target")
+			err := startTarget("", []string{"--logflag all", "2>&1 | tee /tmp/spdk_tgt.log"}, execute)
```
innobead, line 108: Why is GetSPDKDir not required anymore?

innobead (changes requested):

pkg/spdk_test.go
derekbit: In the dapper environment, we already put spdk_tgt in the /user/loca/bin/ which is in the `PATH`, so we don't need to specify `SPDKDir` anymore.

pkg/spdk_test.go
derekbit: Just know this. Thanks for the hint. Updated. Thank you.

innobead (approved):

PR #330, opened 2025-03-17
fix: fix v2 snapshot attribute key name checksum conflict

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10399

change the attr key name used by backing image
we should add a longhorn prefix to the preserved attr name in spdk


shuo-wu (approved):

derekbit (approved):
LGTM

PR #335, opened 2025-04-02
fix(test): fix device busy error caused by in-progress checksum calculation

derekbit:

Issue longhorn/longhorn#10622


pkg/spdk_test.go
```diff
@@ -1234,8 +1534,17 @@ func (s *TestSuite) TestSPDKMultipleThreadSnapshotOpsAndRebuilding(c *C) {
 	}
 }
 
-func (s *TestSuite) TestSPDKMultipleThreadFastRebuilding(c *C) {
```
DamiaSan, line 1237: Is this refactoring, and all the work done over the backing image, related to the _device busy error caused by in-progress checksum calculation_?

derekbit: The PR does
- Separate test cases with backing image and without backing image
- Others are for addressing device busy error caused by in-progress checksum calculation

DamiaSan (approved):

PR #337, opened 2025-04-07
test: only one SPDK CI test is allowed at a time by using a file lock

derekbit:

Issue Longhorn/longhorn#10684

During CI testing, spdk_tgt is repeatedly started and stopped. To prevent interference between concurrent CI tasks, only one SPDK CI test is allowed at a time by using a file lock.

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -1,6 +1,19 @@
 #!/bin/bash
 set -e
 
+LOCK_FILE="/tmp/longhorn-spdk.lock"
+MAX_WAIT=36000  # seconds
+
+# Try to acquire exclusive lock on $LOCK_FILE using file descriptor 200
+# If the lock is held by another process, wait and retry
+# Try to acquire exclusive lock on $LOCK_FILE using file descriptor 200
+# Wait up to $MAX_WAIT seconds
+exec 200>"$LOCK_FILE"
+if ! flock -w "$MAX_WAIT" 200; then
+  echo "Failed to acquire lock on $LOCK_FILE after $MAX_WAIT seconds. Exiting."
+  exit 1
+fi
+
 cd "$(dirname $0)"/..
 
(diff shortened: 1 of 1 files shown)


DamiaSan:
Just a question: do we have multiple processes that execute concurrently `test` script?

scripts/test
```diff
@@ -36,4 +41,17 @@ PACKAGES="$(find . -name '*.go' -print0 | xargs -0 -I{} dirname {} |  cut -f2 -d
 
 trap "rm -f /tmp/test-disk" EXIT
 
+# Try to acquire exclusive lock on $LOCK_FILE using file descriptor 200
+# If the lock is held by another process, wait and retry
+exec 200>"$LOCK_FILE"
+until flock -n 200; do
```
c3y1huang, line 47: NIT: Consider simplifying the retry logic with `flock -w`. That said, I am also fine with the current.

c3y1huang (dismissed):

DamiaSan:
> > Just a question: do we have multiple processes that execute concurrently the `test` script?
> 
> No, https://github.com/longhorn/longhorn-spdk-engine/blob/main/scripts/test#L39 sets parallel to 1

So what is the utility of the lock you have just inserted?

scripts/test
derekbit: Good idea. Updated, thank you.

derekbit:
> > > Just a question: do we have multiple processes that execute concurrently the `test` script?
> > 
> > 
> > No, [`main`/scripts/test#L39](https://github.com/longhorn/longhorn-spdk-engine/blob/main/scripts/test#L39) sets parallel to 1
> 
> So what is the purpose of the lock you have just inserted?

Multiple `build` CI tasks can be executed simultaneously.

c3y1huang (approved):

PR #343, opened 2025-04-10
feat(snapshot): add labels to snapshot creation

DamiaSan:
Issue https://github.com/longhorn/longhorn/issues/9808

Enable the addition of generic labels to a snapshot during its creation.


go.mod
```diff
@@ -63,3 +63,5 @@ require (
 	k8s.io/mount-utils v0.31.3 // indirect
 	k8s.io/utils v0.0.0-20241210054802-24370beab758 // indirect
 )
+
+replace github.com/longhorn/types v0.0.0-20250407153258-74fbf95c1346 => github.com/DamiaSan/types v0.0.0-20250409115014-1b9032629fe1
```
derekbit, line 67: longhorn/types PR is merged. You can update go.mod.

pkg/spdk/engine.go
```diff
@@ -1235,7 +1235,7 @@ func (e *Engine) ReplicaAdd(spdkClient *spdkclient.Client, dstReplicaName, dstRe
 	opts := &api.SnapshotOptions{
 		Timestamp: util.Now(),
 	}
-	updateRequired, replicasErr, engineErr := e.snapshotOperationWithoutLock(spdkClient, replicaClients, snapshotName, SnapshotOperationCreate, opts)
+	updateRequired, replicasErr, engineErr := e.snapshotOperationWithoutLock(spdkClient, replicaClients, snapshotName, SnapshotOperationCreate, opts, nil)
```
derekbit, line 1238: Not need to pass `labels`?

DamiaSan: It is a rebuilding snapshot, not a user created one, so there shouldn't be labels to pass.

derekbit: Does replica rebuilding also restore the labels from the source replica to the destination replica?

pkg/spdk/engine.go
```diff
@@ -1795,7 +1795,7 @@ func (e *Engine) snapshotOperationPreCheckWithoutLock(replicaClients map[string]
 	return snapshotName, nil
 }
 
-func (e *Engine) snapshotOperationWithoutLock(spdkClient *spdkclient.Client, replicaClients map[string]*client.SPDKClient, snapshotName string, snapshotOp SnapshotOperationType, opts any) (updated bool, replicasErr error, engineErr error) {
+func (e *Engine) snapshotOperationWithoutLock(spdkClient *spdkclient.Client, replicaClients map[string]*client.SPDKClient, snapshotName string, snapshotOp SnapshotOperationType, opts any, labels map[string]string) (updated bool, replicasErr error, engineErr error) {
```
shuo-wu, line 1798: I am thinking if there is a better way to define the input of this function. e.g., `opts any...` rather than `opts any, labels map[string]string`.

Or we don't need to retain this general snapshot operation function.

pkg/spdk_test.go
```diff
@@ -326,7 +326,7 @@ func (s *TestSuite) TestSPDKMultipleThread(c *C) {
 			c.Assert(cksumBefore1, Not(Equals), "")
 
 			snapshotName1 := "snap1"
-			_, err = spdkCli.EngineSnapshotCreate(engineName, snapshotName1)
+			_, err = spdkCli.EngineSnapshotCreate(engineName, snapshotName1, nil)
```
shuo-wu, line 329: It's better to simply test this label feature in one of the unit tests

pkg/spdk/engine.go
DamiaSan: No, because actually to retrieve an xattr from SPDK you have to know xattr's name. This is the reason why in [Lvol struct](https://github.com/longhorn/longhorn-spdk-engine/blob/9125fc466faf348115d95064adba6018f615bad6/pkg/api/types.go#L33) there aren't an array for storing general labels, we have only `SnapshotTimestamp` and `UserCreated`, whicha are stored as xattr but they have a fixed and so known name.

DamiaSan: Probably some extra development are needed, if inside Longhorn  the only place where we store information is SPDK, we should need a method to get all the xattrs of a lvol.

derekbit: So, does it mean it is possible the labels are all missing after multiple rebuilding?
If Yes, please go create a BUG ticket for fixing the bug.

pkg/spdk_test.go
DamiaSan: As mentioned above https://github.com/longhorn/longhorn-spdk-engine/pull/343#discussion_r2039142500, actually we can't retrieve general labels, some extra development need to be done.

pkg/spdk/engine.go
DamiaSan: What do you suggest?

pkg/spdk/engine.go
derekbit: > I mean that general labels will be missing also in the original snapshot, the function `BdevLvolGetXattr` is used only for labels with fixed names. So extra development is needed. I think we can track it in the issue I am working on: [longhorn/longhorn#9808](https://github.com/longhorn/longhorn/issues/9808)

Sure, but remember to finish the replica rebuilding APIs first, because it will impact Shuo's integration progress. You can continue the remaining this after implementing those APIs.

pkg/spdk/engine.go
shuo-wu: I mean, can the func definition be like `func (e *Engine) snapshotOperationWithoutLock(spdkClient *spdkclient.Client, replicaClients map[string]*client.SPDKClient, snapshotName string, snapshotOp SnapshotOperationType, opts any...) (updated bool, replicasErr error, engineErr error) {`

pkg/spdk_test.go
shuo-wu: OK. You can test if we can set and retrieve labels with specific keys first. Or add the testing after the extra API is delivered.

derekbit:
@DamiaSan Will you address https://github.com/longhorn/longhorn-spdk-engine/pull/343#discussion_r2039200946 in the PR? If Yes, I will review it once everything is ready.

derekbit:
@DamiaSan Just noticed the PR is closed. Why did you close the PR?

DamiaSan:
> @DamiaSan The PR was being reviewed. Just noticed the PR is closed. Why did you close the PR?

The reason is described here https://github.com/longhorn/longhorn-spdk-engine/pull/343#discussion_r2039200946

derekbit:
> > @DamiaSan The PR was being reviewed. Just noticed the PR is closed. Why did you close the PR?
> 
> The reason is described here [#343 (comment)](https://github.com/longhorn/longhorn-spdk-engine/pull/343#discussion_r2039200946)

Got it. To prevent confusion in issue management, please make sure to update the status before closing a PR or ticket in the future.

derekbit:
> > @DamiaSan The PR was being reviewed. Just noticed the PR is closed. Why did you close the PR?
> 
> The reason is described here [#343 (comment)](https://github.com/longhorn/longhorn-spdk-engine/pull/343#discussion_r2039200946)

Got it. To prevent confusion in issue management, please make sure to update the status or the reason before closing a PR or ticket in the future.

cc @innobead

DamiaSan:
> > > @DamiaSan The PR was being reviewed. Just noticed the PR is closed. Why did you close the PR?
> > 
> > 
> > The reason is described here [#343 (comment)](https://github.com/longhorn/longhorn-spdk-engine/pull/343#discussion_r2039200946)
> 
> Got it. To prevent confusion in issue management, please make sure to update the status or the reason before closing a PR or ticket in the future.
> 
> cc @innobead

To prevent confusion in issue management, please add details about v1 labels management, and not only a general "V2 volume snapshot should support labels". So that a developer, before to start implementing the feature also for v2 engine, could have a clear idea of the jobs to do.
Thanks.

cc @innobead

derekbit:
> > > > @DamiaSan The PR was being reviewed. Just noticed the PR is closed. Why did you close the PR?
> > > 
> > > 
> > > The reason is described here [#343 (comment)](https://github.com/longhorn/longhorn-spdk-engine/pull/343#discussion_r2039200946)
> > 
> > 
> > Got it. To prevent confusion in issue management, please make sure to update the status or the reason before closing a PR or ticket in the future.
> > cc @innobead
> 
> To prevent confusion in issue management, please add details about v1 labels management, and not only a general "V2 volume snapshot should support labels". So that a developer, before to start implementing the feature also for v2 engine, could have a clear idea of the jobs to do. Thanks.
> 
> cc @innobead

The feature is mapped to the v1 volume feature. You can and should check the behavior in Longhorn manager and official docuemnt. 

> > > > @DamiaSan The PR was being reviewed. Just noticed the PR is closed. Why did you close the PR?
> > > 
> > > 
> > > The reason is described here [#343 (comment)](https://github.com/longhorn/longhorn-spdk-engine/pull/343#discussion_r2039200946)
> > 
> > 
> > Got it. To prevent confusion in issue management, please make sure to update the status or the reason before closing a PR or ticket in the future.
> > cc @innobead
> 
> To prevent confusion in issue management, please add details about v1 labels management, and not only a general "V2 volume snapshot should support labels". So that a developer, before to start implementing the feature also for v2 engine, could have a clear idea of the jobs to do. Thanks.
> 
> cc @innobead

TBH, not really interested in debating this. The feature is mapped to the v1 volume feature as we mentioned. Codes and document are there. You can and should check the behavior in codes and official document and make the behavior consistent while implementing the feature. We have mentioned this again and again. It is the responsibility of the issue owner :)

DamiaSan:
> TBH, not really interested in debating this. The feature is mapped to the v1 volume feature as we mentioned. Codes and document are there. You can and should check the behavior in codes and official document and make the behavior consistent while implementing the feature. We have mentioned this again and again. It is the responsibility of the issue owner :)

You are not interested in debating this, but you are still writing about your reasons.

PR #349, opened 2025-04-18
fix: dont delete backing image when rebuilding

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10527

`GetBdevLvolMapWithFilter(spdkClient, r.replicaLvolFilter)` will get the backing image as well,
**This is because, in some of the verify logic, we will need to get the backing image here to construct the chain with the backing image lvol, we can't only get the snapshot of the replica. The chain needs to include the backing image.**

```
func (r *Replica) replicaLvolFilter(bdev *spdktypes.BdevInfo) bool {
	if bdev == nil || len(bdev.Aliases) < 1 || bdev.DriverSpecific.Lvol == nil {
		return false
	}
	lvolName := spdktypes.GetLvolNameFromAlias(bdev.Aliases[0])
	// it is okay to have backing image snapshot in the results, because we exclude it when finding root or construct the snapshot map
	return IsReplicaLvol(r.Name, lvolName) || IsBackingImageSnapLvolName(lvolName)
}
```

Thus, in some of the logic, we should skip the backing image lvol when iterating the map from `GetBdevLvolMapWithFilter()`
In this case, we shouldn't delete the backing image, it is not related to the rebuild

### DIFF
--- a/pkg/spdk/replica.go
+++ b/pkg/spdk/replica.go
@@ -1864,7 +1864,10 @@ func (r *Replica) RebuildingDstStart(spdkClient *spdkclient.Client, srcReplicaNa
 		return "", err
 	}
 	for lvolName, lvol := range bdevLvolMap {
-		if lvolName == r.Name || (r.ActiveChain[0] != nil && lvolName == r.ActiveChain[0].Name) {
+		if IsBackingImageSnapLvolName(lvolName) {
+			continue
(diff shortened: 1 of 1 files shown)


derekbit (approved):
LGTM

derekbit (approved):

PR #357, opened 2025-05-06
fix: parent field of a backing image returned to the instance manager should be empty

derekbit:

Issue Longhorn/longhorn#10526

The parent field of a backing image returned to the instance manager should be empty.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -47,12 +47,16 @@ func ProtoLvolToLvol(l *spdkrpc.Lvol) *Lvol {
 	if l == nil {
 		return nil
 	}
+	parent := l.Parent
+	if types.IsBackingImageSnapLvolName(parent) {
+		parent = ""
+	}
 	return &Lvol{
 		Name: l.Name,
 		// UUID:         l.Uuid,
 		SpecSize:          l.SpecSize,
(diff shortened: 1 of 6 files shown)


c3y1huang (approved):

pkg/api/types.go
```diff
@@ -47,12 +47,16 @@ func ProtoLvolToLvol(l *spdkrpc.Lvol) *Lvol {
 	if l == nil {
 		return nil
 	}
+	parent := l.Parent
+	if types.IsBackingImageSnapLvolName(parent) {
```
shuo-wu, line 51: Should this be `if types.IsBackingImageSnapLvolName(l.Name)`?

derekbit: `l.Name` is the name of the replica that is a child of the backing image, and `l.Parent` refers to the `base_snapshot`, which is the name of the backing image.

For v1 volumes, the replica that immediately follows the backing image should have an empty `Parent` field. Therefore, if the `base_snapshot` is the backing image, I set `l.Parent` to an empty string.

shuo-wu: Got it. I was thinking that this PR will unset `parent` field of a backing image lvol.

shuo-wu (approved):

PR #363, opened 2025-05-14
feat: return UUID in engine and replica Get and List operations

derekbit:

Issue Longhorn/longhorn#10888


pkg/api/types.go
```diff
@@ -130,6 +132,7 @@ func ReplicaToProtoReplica(r *Replica) *spdkrpc.Replica {
 		Rebuilding: r.Rebuilding,
 		State:      r.State,
 		ErrorMsg:   r.ErrorMsg,
+		Uuid:       r.UUID,
```
derekbit, line 135: nit: You have one commit for `sync uuid to Engine from rpc call`. It would be better if you can move the changes in `ReplicaToProtoReplica` and `ProtoReplicaToReplica` to a new commit for the consistency.

COLDTURNIP:
Wrong base to the last update, fixing.

c3y1huang (commented):
Commit-lint failed.

pkg/api/types.go
```diff
@@ -98,6 +99,7 @@ func ProtoReplicaToReplica(r *spdkrpc.Replica) *Replica {
 		State:      r.State,
 		ErrorMsg:   r.ErrorMsg,
 		Rebuilding: r.Rebuilding,
+		UUID:       r.Uuid,
```
c3y1huang, line 102: What is the difference between LvsUUID and UUID?

derekbit: LvsUUID is the UUID of the logical volume store (LVS). UUID is for the logical volume (LV). LVS is a storage pool for LVs.

c3y1huang (approved):

COLDTURNIP (approved):

PR #364, opened 2025-05-15
fix: rebuilding replica ancestor snapshot will miss backing image

shuo-wu:

Longhorn/longhorn#10909

Since the rebuilding snapshot list provided by the src replica will not tell if the ancestor snapshot has the backing image as its parent.

### DIFF
--- a/pkg/spdk/replica.go
+++ b/pkg/spdk/replica.go
@@ -2159,26 +2159,14 @@ func (r *Replica) rebuildingDstShallowCopyPrepare(spdkClient *spdkclient.Client,
 	if srcSnapSvcLvol == nil {
 		return "", fmt.Errorf("cannot find snapshot %s in the rebuilding snapshot list for replica %s shallow copy prepare", snapshotName, r.Name)
 	}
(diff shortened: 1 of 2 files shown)


pkg/spdk/replica.go
```diff
@@ -2229,7 +2217,7 @@ func (r *Replica) rebuildingDstShallowCopyPrepare(spdkClient *spdkclient.Client,
 			r.log.Infof("Replica found an intact snapshot lvol %s before the shallow copy", dstSnapSvcLvol.Alias)
 			rebuildingLvolCreated = true
 		}
-	} else { // Then check if there is an expired lvol available
+	} else {                                               // Then check if there is an expired lvol available
```
derekbit, line 2220: go fmt failed

derekbit (dismissed):
LGTM. Only need to fix the go fmt error.

derekbit:
Build is passed in https://github.com/longhorn/longhorn-spdk-engine/actions/runs/15057836568?pr=364

derekbit (approved):

PR #366, opened 2025-05-19
feat: v2 volume range hash

shuo-wu:

Issue longhorn/longhorn#10037

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -238,6 +238,23 @@ func (c *SPDKClient) ReplicaSnapshotHashStatus(name, snapshotName string) (*spdk
 	})
 }
 
+func (c *SPDKClient) ReplicaSnapshotRangeHashGet(name, snapshotName string, clusterStartIndex, clusterCount uint64) (*spdkrpc.ReplicaSnapshotRangeHashGetResponse, error) {
+	if name == "" || snapshotName == "" {
+		return nil, fmt.Errorf("failed to get range hash for SPDK replica snapshot: missing required parameter name or snapshot name")
+	}
+
+	client := c.getSPDKServiceClient()
+	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceTimeout)
+	defer cancel()
+
+	return client.ReplicaSnapshotRangeHashGet(ctx, &spdkrpc.ReplicaSnapshotRangeHashGetRequest{
+		Name:              name,
+		SnapshotName:      snapshotName,
+		ClusterStartIndex: clusterStartIndex,
+		ClusterCount:      clusterCount,
+	})
+}
+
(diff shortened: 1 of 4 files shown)


DamiaSan (commented):
LGTM, will approve once the PR is finalized

shuo-wu:
It seems that the unit test won't apply the latest spdk then get failed without changing https://github.com/longhorn/longhorn-spdk-engine/blob/1ca092276c9732bf3b390905d4b5347e7acad854/Dockerfile.dapper#L47, even though https://github.com/longhorn/dep-versions is already updated

shuo-wu:
With the latest SPDK commit, both unit tests and the manual test get passed.

shuo-wu:
> It seems that the unit test won't apply the latest spdk then get failed without changing

## RCA:

The latest commit of https://github.com/longhorn/dep-versions is not tagged. Then in this repo CI stage, [the CI will switch to the more recent **tagging** version](https://github.com/longhorn/longhorn-spdk-engine/blob/7f7068a9acea1aa2ecf4f6da860f0e2f9fbb5f51/Dockerfile.dapper#L38-L48) `3028cc58ac215768d3b89298d63d38564d7d6f95 (HEAD, tag: v24.09.0+4)` **rather than the latest commit** `68191306ab3d3f30432bcbf55189592d47d8f510` for longhorn/dep-versions, which means some SPDK fixes not included:
```
    if [ -n "${SRC_TAG}" ] && git show-ref --tags ${SRC_TAG} > /dev/null 2>&1; then \
        echo "Checking out tag ${SRC_TAG}"; \
        cd /usr/src/dep-versions && git checkout tags/${SRC_TAG}; \
    fi
```

Actually, I am thinking whether this tag switching logic is really necessary.

shuo-wu:
YES. The weird thing is, `SRC_TAG` should be empty and the `if` code block should not take any effect. But the result is, Dockerfile.dapper will switch to commit `18681b2319b0cbd793a5d216eece8242b2329883 fix(workflow): use longhorn bot as author and committer` for /usr/src/dep-versions after adding that `if` code block

derekbit:
> YES. The weird thing is, `SRC_TAG` should be empty and the `if` code block should not take any effect. But the result is Dockerfile.dapper will use commit `18681b2319b0cbd793a5d216eece8242b2329883 fix(workflow): use longhorn bot as author and committer` for /usr/src/dep-versions after adding that `if` code block.
> 
> I am not sure if it's the cache of my personal dev machine that leads to this case. I tried to clean up all docker data via `docker system prune` then re-run `make`. But this does not work.
> 
> If possible, would you give it a try in your dev env? Besides, this PR itself is ready for review

Fixed. It is caused by the docker cache. I've added the cache buster in the dockerfile.dapper https://github.com/longhorn/longhorn-spdk-engine/pull/373. You can rebase the latest codes and give it a try.

derekbit (approved):
LGTM

PR #376, opened 2025-06-17
[WIP] feat(engine): add I/O suspension during replica deletion

hookak:

https://github.com/longhorn/longhorn/issues/11109


pkg/spdk/engine.go
```diff
@@ -1678,6 +1678,19 @@ func (e *Engine) ReplicaDelete(spdkClient *spdkclient.Client, replicaName, repli
 		return fmt.Errorf("replica %s recorded address %s does not match the input address %s for engine %s replica delete", replicaName, replicaStatus.Address, replicaAddress, e.Name)
 	}
 
+	if e.Frontend == types.FrontendSPDKTCPBlockdev && e.Endpoint != "" {
+		if err = e.initiator.Suspend(false, false); err != nil {
```
shuo-wu, line 1682: I am thinking if suspending with flag `noflush` would help completely avoid pending I/Os in the `io_submitted` list

hookak: Sorry for the delayed response. @shuo-wu 

First, when I set `noflush` to `true`, the issue still occurred.
(I believe it’s also worth running repeated tests with `noflush` set to `false`.)

What I’m currently looking into is that, when the RAID bdev removes a base bdev(`spdkClient.BdevRaidRemoveBaseBdev`), SPDK internally calls `spdk_bdev_quiesce()`, which also functions to pause I/O. After that, the base bdev is removed, but at that moment the `io_submitted` list of the base bdev is not empty, causing the issue.

Therefore, I’m checking whether this can be addressed within SPDK. If that turns out to be difficult, I think it would be better to keep the current approach of using `dmsetup` suspend.

What do you think?

shuo-wu (approved):

innobead:
cc @longhorn/maintainer

hookak:
Thanks for testing this PR, @bachmanity1 - much appreciated. I agree the patch didn't resolve the issue. This looks like something that should be fixed in SPDK itself. Would it be OK to revert this patch from `master` branch for now?

PR #379, opened 2025-07-01
fix: server should scan possible existing replicas when adding a new disk

shuo-wu:

Issue
https://github.com/longhorn/longhorn/issues/10828
https://github.com/longhorn/longhorn/issues/11188


shuo-wu:
The e2e test https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3459/ got passed

derekbit (approved):
LGTM

PR #385, opened 2025-07-22
feat: support interrupt mode for NVMe/TCP

c3y1huang:

Issue longhorn/longhorn#9834, ~longhorn/longhorn#9419~

Support interrupt mode for NVMe/TCP


c3y1huang:
@coderabbitai review

PR #386, opened 2025-07-25
feat(engine, replica): v2 volume expand #8022

davidcheng0922:

https://github.com/longhorn/longhorn/issues/8022

support v2 volume expansion, only for nvme frontend

[v2-expansion.webm](https://github.com/user-attachments/assets/229db8a5-d084-4b4f-9048-139b7f4ec281)
[v2-expansion-with-live-IO.webm](https://github.com/user-attachments/assets/95cfe596-e33d-418d-837c-3cae69faf18e)


pkg/spdk/engine.go
```diff
@@ -1229,6 +1304,356 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	}
+
+	if e.IsRestoring {
+		return fmt.Errorf("restoring is in progress")
+	}
+
+	if types.IsUblkFrontend(e.Frontend) {
+		return errors.New("ublk frontend not support expansion")
+	}
+
```
PhanLe1010, line 1421: Do we have a ticket to improve this?

davidcheng0922: Thanks for remind, I raise the ticket for it
https://github.com/longhorn/longhorn/issues/11406

pkg/spdk/engine.go
```diff
@@ -1229,6 +1304,388 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+		e.log.Infof("Replica had been expanded to size %v", size)
+		return nil
+	}
+
+	roundedNewSize := util.RoundUp(size, helpertypes.MiB)
+	if roundedNewSize != size {
+		return fmt.Errorf("rounded up spec size from %v to %v since the spec size should be multiple of MiB", size, roundedNewSize)
+	}
+
+	e.isExpanding = true
```
shuo-wu, line 1444: The size check should be executed before the `defer` error handling, since:
1. A wrong expansion request should not error the engine
2. Retry for an invalid size expansion does not make sense.

pkg/spdk/engine.go
```diff
@@ -1229,6 +1304,388 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	if err != nil {
+		return err
+	}
+	defer e.closeRplicaClients(replicaClients)
+	if len(replicaClients) == 0 || e.isAnyReplicaUnderRebuilding(replicaClients) {
+		return errors.New("expansion failed: no healthy replica or at least one replica is under rebuilding")
+	}
+
+	// prepareRaidForExpansion checks if RAID exists, suspends frontend if needed, and deletes the RAID bdev.
+	isSusupend, bdevRaidUUID, err := e.prepareRaidForExpansion(spdkClient)
```
shuo-wu, line 1371: The returned error of this func is probably an engine error, which is typically related to the frontend or RAID and probably affects the IO. For this kind of error, we need to crash the engine.

pkg/spdk/engine.go
```diff
@@ -1229,6 +1304,388 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	if len(bdevRaid) == 0 {
+		return false, "", nil // RAID already deleted
+	}
+	bdevUUID = bdevRaid[0].UUID
+
+	// Suspend IO if frontend is active
+	if e.Endpoint != "" && e.initiator != nil {
+		if err := e.initiator.Suspend(true, true); err != nil {
+			return false, "", errors.Wrapf(err, "failed to suspend initiator for engine %s", e.Name)
+		}
```
shuo-wu, line 1498: As we discussed in the LEP meeting, this `initiator.Suspend` happens in device mapper layer, and `initiator.replaceDmDeviceTarget` can reload and resume the underlying block device without worrying about the previous device. This means that we can directly delete both the initiator and the target of the frontend, no matter what kind of frontend the engine is using.
In brief, we may be able to use a general flow to handling both nvmf and ublk expansion

pkg/spdk/engine.go
```diff
@@ -1229,6 +1304,388 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+		// all success
+		e.log.Warn("All replicas expand success")
+		return true
+	case len(failedReplica) == len(e.ReplicaStatusMap):
+		// all fail, no based lvols are resized
+		e.log.WithFields(logrus.Fields{
+			"engineName": e.Name,
+			"volumeName": e.VolumeName,
+		}).Warn("All replicas failed to expand; proceeding with RAID recreation using existing size")
+		return false
```
shuo-wu, line 1611: Can we aggregate errors and return the error here? As we discussed, if all replicas fail, it's better to crash the engine and let the salvage feature help recovery.

pkg/spdk/engine.go
```diff
@@ -1229,6 +1304,388 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+
+	// recreate RAID and reconnect frontend
+	if err := e.reconnectFrontend(spdkClient, bdevRaidUUID); err != nil {
+		e.log.WithFields(logrus.Fields{
+			"engineName": e.Name,
+			"volumeName": e.VolumeName,
+			"error":      err,
+		}).Error("reconnectFrontend failed")
+		return err
+	}
```
shuo-wu, line 1395: Do we need to recreate the RAID if `expansionSucceeded == false`?

pkg/spdk/engine.go
davidcheng0922: Quick question for this, 
Yeah, it seems that `initiator.Suspend` is already a general suspend for nvmf and ublk in dm layer.
What else should I do for `Suspend`? 

and I plan to replace `e.initiator.Resume()` with `e,inititator.replaceDmDeviceTarget`

davidcheng0922: **Ublk**
- `dmsetup suspend` waits for all in-flight I/O to complete and blocks new I/O.
- With a ublk device, all I/O is processed in `userspace`
    - If the ublk userspace process is idle, blocked, or terminated, then:
    - Kernel I/O queues stall.
    - dmsetup suspend waits forever for those I/Os to complete. It will be silent stuck
    
 **NVMe**
nvme will not have this issue, once `nvme connect`, it becomes kernel block device. I/O are handled by NVMe driver in Linux kernel.

pkg/spdk/engine.go
```diff
@@ -862,7 +930,19 @@ func (e *Engine) ValidateAndUpdate(spdkClient *spdkclient.Client) (err error) {
 
 	bdevRaidSize := bdevRaid.NumBlocks * uint64(bdevRaid.BlockSize)
 	if e.SpecSize != bdevRaidSize {
-		return fmt.Errorf("found mismatching between engine spec size %d and actual raid bdev size %d for engine %s", e.SpecSize, bdevRaidSize, e.Name)
+		// not directly return error
+
+		// If the volume is not attached and do the expand
+		// At first, we create and attach the engine with new size, but not yet to expand
```
shuo-wu, line 936: Do we need to do anything for `EngineCreate` with a larger spec size?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1309,379 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+
+	// fetch all replica clients
+	replicaClients, err := e.getReplicaClients()
+	if err != nil {
+		return err
+	}
+	defer e.closeRplicaClients(replicaClients)
+	if len(replicaClients) == 0 || e.isAnyReplicaUnderRebuilding(replicaClients) {
+		return errors.New("expansion failed: no healthy replica or at least one replica is under rebuilding")
+	}
```
shuo-wu, line 1367: Can we directly check `e.ReplicaStatusMap` inside `startExpansion()` instead?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1309,379 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+}
+
+func (e *Engine) finishExpansion(expanded bool, size uint64) {
+	if expanded {
+		if e.lastExpansionError != "" {
+			logrus.Infof("Succeeded to expand from size %v to %v but there are some replica expansion failures: %v", e.SpecSize, size, e.lastExpansionError)
+		} else {
+			logrus.Infof("Succeeded to expand from size %v to %v", e.SpecSize, size)
+		}
+		e.SpecSize = size
```
shuo-wu, line 1444: Use `e.log` instead. And need to update `specSize` for `e.log`

pkg/spdk/engine.go
```diff
@@ -1229,6 +1309,379 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+		}
+		return false, "", errors.Wrap(err, "failed to get bdev raid")
+	}
+	if len(bdevRaid) == 0 {
+		return false, "", nil // RAID already deleted
+	}
+	bdevUUID = bdevRaid[0].UUID
+
+	// Suspend IO if frontend is active
+	if e.Frontend == types.FrontendSPDKTCPBlockdev && e.Endpoint != "" {
```
shuo-wu, line 1482: For now, need to return an error when the frontend is ublk

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,91 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+	// resize it if not equal
+	if bdevLvol.SpecSize != size {
+		resized, err := spdkClient.BdevLvolResize(r.Alias, util.BytesToMiB(size))
+		if err != nil {
+			r.log.Errorf("Resize replica %s failed, %v", r.Name, err)
+			return errors.Wrapf(err, "bdev lvol resize error")
+		} else if !resized {
+			return fmt.Errorf("no error, but replica %s not resized", r.Name)
+		}
+	}
```
shuo-wu, line 1086: Do we need to redo the expose?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1333,399 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	}
+
+	if e.IsRestoring {
+		return fmt.Errorf("restoring is in progress")
+	}
+
+	if e.SpecSize > size {
+		return fmt.Errorf("cannot expand engine to a smaller size %v, current size %v", size, e.SpecSize)
+	} else if e.SpecSize == size {
+		e.log.Infof("Replica had been expanded to size %v", size)
```
PhanLe1010, line 1446: Should we make sure that the caller Expand() also return immediately instead of continue to perform the expansion?

PhanLe1010, line 1446: Should we make sure that the caller Expand() also return immediately instead of continue to perform the expansion?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1333,399 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	if len(e.ReplicaStatusMap) == 0 {
+		return fmt.Errorf("cannot expand engine with no replica")
+	}
+
+	// at least one healthy replica is required for expansion
+	containHealthyReplica := false
+	for _, replicaStatus := range e.ReplicaStatusMap {
+		if replicaStatus.Mode != types.ModeERR {
+			containHealthyReplica = true
+			break
```
PhanLe1010, line 1464: I think healthy replica should be `replicaStatus.Mode == types.ModeRW`.

Should we only start the expansion if ALL replica is healthy instead of at least 1 replica is healthy

pkg/spdk/engine.go
```diff
@@ -1229,6 +1333,399 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+		if engineErr != nil {
+			if e.State != types.InstanceStateError {
+				e.State = types.InstanceStateError
+			}
+			e.ErrorMsg = engineErr.Error()
+
+			if errUpdateLogger := e.log.UpdateLogger(logrus.Fields{
+				"replicaStatusMap": e.ReplicaStatusMap,
+			}); errUpdateLogger != nil {
+				e.log.WithError(errUpdateLogger).Warn("Failed to update logger with replica status map during engine creation")
```
PhanLe1010, line 1373: Is this needed?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1333,399 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+			e.log.WithError(engineErr).Errorf("Engine %s failed to expand", e.Name)
+			e.lastExpansionError = errors.Wrap(retErr, "engine failed to expand expansion").Error()
+			e.lastExpansionFailedAt = time.Now().UTC().Format(time.RFC3339Nano)
+		}
+
+		if engineErr != nil {
+			if e.State != types.InstanceStateError {
+				e.State = types.InstanceStateError
+			}
+			e.ErrorMsg = engineErr.Error()
```
PhanLe1010, line 1368: Should we print out a log here?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1333,399 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	replicaClients, err := e.getReplicaClients()
+	if err != nil {
+		retErr = err
+		return retErr
+	}
+	defer e.closeRplicaClients(replicaClients)
+	if e.isAnyReplicaUnderRebuilding(replicaClients) {
+		retErr = errors.New("expansion failed: at least one replica is under rebuilding")
+		return retErr
+	}
```
PhanLe1010, line 1393: rebuilding replica can be detected by `replicaStatus.Mode != types.ModeRW` above

pkg/spdk/engine.go
```diff
@@ -1229,6 +1333,399 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+		return suspendFrontend, "", errors.Wrap(err, "failed to get bdev raid")
+	}
+	if len(bdevRaid) == 0 {
+		return suspendFrontend, "", nil // RAID already deleted
+	}
+	bdevUUID = bdevRaid[0].UUID
+
+	// Suspend IO if frontend is active
+	if e.Frontend == types.FrontendSPDKTCPBlockdev && e.Endpoint != "" {
+		if err := e.initiator.Suspend(false, false); err != nil {
```
PhanLe1010, line 1525: Just want to double check. This function means dm device suspension, right?

davidcheng0922: yeah, it actually calls 
`dmsetup suspend --noflush=false --nolockfs=false`

pkg/client/client.go
```diff
@@ -111,6 +111,22 @@ func (c *SPDKClient) ReplicaGet(name string) (*api.Replica, error) {
...
+
+	client := c.getSPDKServiceClient()
+	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceTimeout)
+	defer cancel()
+
+	_, err := client.ReplicaExpand(ctx, &spdkrpc.ReplicaExpandRequest{
+		Name: name,
+		Size: size,
+	})
+	return errors.Wrapf(err, "failed to expand SPDK replica %v", name)
```
derekbit, line 127:
```suggestion
	return errors.Wrapf(err, "failed to expand replica %v", name)
```

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,110 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+		return err
+	}
+
+	roundedSize := util.RoundUp(size, clusterSize)
+	if roundedSize != size {
+		return fmt.Errorf("replica %s rounded up spec size from %v to %v since the spec size should be multiple of MiB", r.Name, size, roundedSize)
+	}
+
+	if r.SpecSize > roundedSize {
+		return fmt.Errorf("cannot expand replica %s to a smaller size %v, current spec size %v", r.Name, size, r.SpecSize)
```
derekbit, line 1063: Curious: Can SPDK support shrink the size of a logical volume?

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,110 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+		reExposeBdev := false
+		if r.IsExposed {
+			if err := spdkClient.StopExposeBdev(helpertypes.GetNQN(r.Name)); err != nil && !jsonrpc.IsJSONRPCRespErrorNoSuchDevice(err) {
+				return fmt.Errorf("failed to stop expose replica %v before expansion: %w", r.Name, err)
+			}
+			r.IsExposed = false
+			reExposeBdev = true
+		}
+
+		resized, err := spdkClient.BdevLvolResize(r.Alias, util.BytesToMiB(size))
```
derekbit, line 1089: It only expands the volume head. Should we handle the snapshots?

pkg/spdk/replica.go
davidcheng0922: Yes, it support shrink, even the data is allocated. The allocated data may be removed directly.

pkg/client/client.go
```diff
@@ -111,6 +111,22 @@ func (c *SPDKClient) ReplicaGet(name string) (*api.Replica, error) {
 	return api.ProtoReplicaToReplica(resp), nil
 }
 
+func (c *SPDKClient) ReplicaExpand(name string, size uint64) error {
+	if name == "" {
+		return fmt.Errorf("failed to delete SPDK replica: missing required parameter")
+	}
+
+	client := c.getSPDKServiceClient()
+	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceTimeout)
```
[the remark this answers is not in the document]
davidcheng0922: On my current understanding, the expansion time will not be affected by the change of size.  We use thin volume and `bdev_lvol_resize` update the metadata, not the actual data allocation
My test from: 1G -> 2G and 1G ->100G are almost the same, both complete instantly.

pkg/spdk/engine.go
```diff
@@ -227,6 +231,26 @@ func (e *Engine) Create(spdkClient *spdkclient.Client, replicaAddressMap map[str
...
+			return nil, errors.Wrapf(err, "failed to get replica %v from %v during engine creation", replicaName, replicaAddr)
+		}
+
+		if e.SpecSize != replica.SpecSize {
+			// not return error here
+			// it may cause infinite retry if the user doesn't fix the size issue
+			e.SpecSize = replica.SpecSize
+			e.log.Warnf("Engine spec size is not consistent with replica %v spec size, will use the replica spec size %v", replicaName, replica.SpecSize)
+			break
+		}
```
shuo-wu, line 251: Make sure all replicas have the same spec size first. And it should return an error if `e.SpecSize < replica.SpecSize`

pkg/spdk/engine.go
```diff
@@ -634,6 +658,67 @@ func (e *Engine) handleUblkFrontend(spdkClient *spdkclient.Client) (err error) {
...
+			}
+			e.log.Infof("Finished reconnecting frontend for engine: %+v", e)
+		}
+	}()
+
+	// Stop previous target (if it exists)
+	// This ensures that any previous exported bdev target with the same NQN is cleanly removed
+	// from the SPDK target subsystem. SPDK does not allow re-exporting a bdev with the same NQN
+	// without first removing the old instance. Skipping this step may result in:
+	// Therefore, this is a defensive cleanup step to make the target namespace idempotent.
```
shuo-wu, line 691: This explains why we need to stop the old nvmf target. But it is not related to why the nvmf target is stopped before the nvmf initiator disconnection. 
Or in other words, can we do something like: `e.initiator.DisconnectNVMeTCPTarget` -> `spdkClient.StopExposeBdev` -> `spdkClient.StartExposeBdev` -> `e.initiator.StartNvmeTCPInitiator`

pkg/spdk/engine.go
```diff
@@ -673,6 +758,7 @@ func (e *Engine) Delete(spdkClient *spdkclient.Client, superiorPortAllocator *co
 	// Stop the frontend
 	if e.initiator != nil {
 		if _, err := e.initiator.Stop(spdkClient, true, true, true); err != nil {
+			e.log.Error(err)
```
shuo-wu, line 761: NIT: You can directly add an error log printing in the `defer` func

pkg/spdk/replica.go
davidcheng0922: I think no need. The previous snapshots are RO and stay their original size.
When we cloning from these snapshots, it will be original size.

pkg/spdk/replica.go
derekbit: Got it. Thanks.

pkg/util/util.go
```diff
@@ -284,3 +284,28 @@ func GetFileChecksum(filePath string) (string, error) {
 
 	return hex.EncodeToString(h.Sum(nil)), nil
 }
+
+func BackoffRetry(attempts int, baseDelay, maxDelay time.Duration, exponentialBackoff float64, backoffFunc func() error) error {
```
derekbit, line 288: Do we need this function?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1281,345 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+		for replicaName, err := range failedReplica {
+			e.ReplicaStatusMap[replicaName].Mode = types.ModeERR
+			aggregatedErr[replicaName] = err.Error()
+		}
+
+		e.log.WithFields(logrus.Fields{
+			"engineName":     e.Name,
+			"volumeName":     e.VolumeName,
+			"failedReplicas": aggregatedErr,
+		}).Error("Some replicas failed to expand and have been marked as ERR")
```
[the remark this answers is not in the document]
davidcheng0922: The dangerous means if we allow shrink in some cases, I'm afraid the data will be erased and the fs is broken.
If we can make sure the data is not used, it might be ok.

pkg/util/util.go
davidcheng0922: I used it in 

[`pkg/spdk/engine.go` lines 1697–1757](https://github.com/longhorn/longhorn-spdk-engine/pull/386/files#diff-2f9990d3c794aaa008aacf70ecd9ae72fc65febf9eee3ea038463cc542e43c36R1697-R1757)

to make sure the raid is created

pkg/spdk/engine.go
```diff
@@ -1229,6 +1346,415 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	// engineErr will be set when the engine failed to do any non-recoverable operations.
+	defer func() {
+		if retErr != nil {
+			e.log.WithError(engineErr).Errorf("Engine %s failed to expand", e.Name)
+			e.lastExpansionError = errors.Wrap(retErr, "engine failed to expand expansion").Error()
+			e.lastExpansionFailedAt = time.Now().UTC().Format(time.RFC3339Nano)
+		}
+
+		if engineErr != nil {
+			if e.State != types.InstanceStateError {
```
innobead, line 1393: To simplify the code, do the assignment statement w/o the expression condition check.

```go
e.State = types.InstanceStateError
```

pkg/spdk/engine.go
```diff
@@ -1229,6 +1346,415 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+				e.State = types.InstanceStateError
+			}
+			e.ErrorMsg = engineErr.Error()
+			e.log.WithError(engineErr).Errorf("Engine %s under non-recoverable operation during expansion", e.Name)
+
+			if errUpdateLogger := e.log.UpdateLogger(logrus.Fields{
```
innobead, line 1399: nit: See this logger update logic repeatedly but the only variable part is `logrus.Fields`. Suggest abstracting this a util function/method instead.

pkg/spdk/engine.go
```diff
@@ -1229,6 +1346,415 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	if err != nil {
+		engineErr = errors.Wrap(err, "prepare raid for expansion failed")
+		retErr = engineErr
+		return retErr
+	}
+
+	// expand replicas
+	if err := e.expandReplicas(replicaClients, spdkClient, size); err != nil {
+		engineErr = err
+		retErr = err
```
innobead, line 1435: nit: `retErr` is already a named returned variable of the method, so you may not need this assignment statement? Revisit other parts as well.

pkg/spdk/engine.go
```diff
@@ -1229,6 +1346,415 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+
+	// recreate RAID and reconnect frontend
+	if err := e.reconnectFrontend(spdkClient, bdevRaidUUID); err != nil {
+		engineErr = errors.Wrap(err, "reconnectFrontend failed")
+		retErr = engineErr
+		return retErr
+	}
+	e.log.Info("Expanding engine complete")
+
+	expanded = true // which could be true even in partial success
```
innobead, line 1447: What does `// which could be true even in partial success` mean? Can't we guarantee expansion is finished in each replica?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1346,415 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+func (e *Engine) startExpansion(size uint64, replicaClients map[string]*client.SPDKClient) (requireExpansion bool, err error) {
+	if e.isExpanding {
+		return false, fmt.Errorf("expansion is in progress")
+	}
+
+	if e.IsRestoring {
+		return false, fmt.Errorf("restoring is in progress")
+	}
+
+	if e.SpecSize > size {
```
innobead, line 1460: Change to below to avoid unnecessary `else` for simplifying code readability.

```go
	if e.SpecSize > size {
		return false, fmt.Errorf("cannot expand engine to a smaller size %v, current size %v", size, e.SpecSize)
	}
    if e.SpecSize == size {
		e.log.Infof("Engine already at requested size %v, skipping expansion", size)
		return false, nil // no need to expand
	}
```

pkg/spdk/engine.go
```diff
@@ -1229,6 +1346,415 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+		replicaClient, ok := replicaClients[replicaName]
+		if !ok {
+			return false, fmt.Errorf("cannot find client for replica %s", replicaName)
+		}
+		replica, err := replicaClient.ReplicaGet(replicaName)
+		if err != nil {
+			return false, errors.Wrapf(err, "cannot get replica %s before expansion", replicaName)
+		}
+
+		if currentReplicaSize == 0 {
```
innobead, line 1493:
```go
		if currentReplicaSize == 0 {
			currentReplicaSize = replica.SpecSize
            continue
		}

        if currentReplicaSize != replica.SpecSize {
			return false, fmt.Errorf("cannot expand engine with replicas in different sizes: replica %s has size %v while other replicas have size %v", replicaName, replica.SpecSize, currentReplicaSize)
		}
```

pkg/spdk/engine.go
```diff
@@ -1229,6 +1346,415 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+		} else if currentReplicaSize != replica.SpecSize {
+			return false, fmt.Errorf("cannot expand engine with replicas in different sizes: replica %s has size %v while other replicas have size %v", replicaName, replica.SpecSize, currentReplicaSize)
+		}
+	}
+
+	if currentReplicaSize > size {
+		return false, fmt.Errorf("cannot expand engine to a smaller size %v, current replica size %v", size, currentReplicaSize)
+	} else if currentReplicaSize == size {
+		e.log.Infof("Replicas already at requested size %v, skipping expansion", size)
+		e.SpecSize = size
```
innobead, line 1504: Why do we need to force change `e.SpecSize`?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1346,415 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	e.lastExpansionFailedAt = ""
+	e.lastExpansionError = ""
+
+	return true, nil
+}
+
+func (e *Engine) finishExpansion(expanded bool, size uint64) {
+	if expanded {
+		if e.lastExpansionError != "" {
+			e.log.Infof("Succeeded to expand from size %v to %v but there are some replica expansion failures: %v", e.SpecSize, size, e.lastExpansionError)
```
innobead, line 1519: If some relicas failed to expand, we still call it a successful expansion? What's the behavior in v1?

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,110 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+	if err != nil {
+		return err
+	}
+
+	roundedSize := util.RoundUp(size, clusterSize)
+	if roundedSize != size {
+		return fmt.Errorf("replica %s rounded up spec size from %v to %v since the spec size should be multiple of MiB", r.Name, size, roundedSize)
+	}
+
+	if r.SpecSize > roundedSize {
```
innobead, line 1062: nit: You may just use `size` directly, as `roundedSize` is just for checking if `size` is equal to `roundedSize`. Using `size` and `roundedSize` in code might cause some confusion.

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,110 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+	// double check if size is already be expanded
+	headBdevLvol, err := spdkClient.BdevLvolGetByName(r.Alias, 0)
+	if err != nil {
+		r.log.Errorf("Get replica %s failed, %v", r.Name, err)
+		return err
+	}
+	bdevLvol := BdevLvolInfoToServiceLvol(&headBdevLvol)
+
+	// resize it if not equal
+	if bdevLvol.SpecSize != size {
```
innobead, line 1078: Do we need to check the valid case here `bdevLvol.SpecSize < Size`? for other cases, should not proceed expansion.

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,110 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+		if r.IsExposed {
+			if err := spdkClient.StopExposeBdev(helpertypes.GetNQN(r.Name)); err != nil && !jsonrpc.IsJSONRPCRespErrorNoSuchDevice(err) {
+				return fmt.Errorf("failed to stop expose replica %v before expansion: %w", r.Name, err)
+			}
+			r.IsExposed = false
+			reExposeBdev = true
+		}
+
+		resized, err := spdkClient.BdevLvolResize(r.Alias, util.BytesToMiB(size))
+		if err != nil {
```
innobead, line 1090: Change to below. Simply the code when dealing with `if else with return`.  

```go
		if err != nil {
			r.log.Errorf("Resize replica %s failed, %v", r.Name, err)
			return errors.Wrapf(err, "bdev lvol resize error")
		}

        if !resized {
			return fmt.Errorf("no error, but replica %s not resized", r.Name)
		}
```

pkg/util/util.go
innobead, line 288: Have you considered using `retry.RetryOnConflict` in k8s.io/client-go/util/retry instead? as go-common-libs depend on that client lib.

innobead (changes requested):

pkg/spdk/engine.go
```diff
@@ -227,6 +231,26 @@
...
+		replicaClient, err := GetServiceClient(replicaAddr)
+		if err != nil {
+			return nil, err
+		}
+		replica, err := replicaClient.ReplicaGet(replicaName)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get replica %v from %v during engine creation", replicaName, replicaAddr)
+		}
+
+		if e.SpecSize != replica.SpecSize {
```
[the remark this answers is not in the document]
davidcheng0922: Just in case.

There is one case I found and resolved. 

If the volume is detached and we expand the volume, we will create the engine first.
The volume will use the new size, and the replica will stay in original size and it will be inconsistent at this moment because the expand operation is not yet executed.

pkg/spdk/engine.go
```diff
@@ -862,7 +967,19 @@ func (e *Engine) ValidateAndUpdate(spdkClient *spdkclient.Client) (err error) {
...
+
+		// If the volume is not attached and do the expand
+		// At first, we create and attach the engine with new size, but not yet to expand
+		// it will cause infinite loop for size mismatching
+		// loop to destroy and create engine
+		// and there is no chance to execute EngineExpand()
+
+		// wait the lh-manager to reconcile engine CR and call EngineExpand()
+
+		e.SpecSize = bdevRaidSize
```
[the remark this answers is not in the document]
davidcheng0922: yeah, you're right

pkg/spdk/engine.go
davidcheng0922: It means we allow partial success which align to v1 expansion logic

pkg/spdk/engine.go
davidcheng0922: yeah, I'll fix this kind of logic

pkg/spdk/engine.go
```diff
@@ -1229,6 +1346,415 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	defer e.closeRplicaClients(replicaClients)
+
+	// startExpansion checks if the expansion can be started, and marks the expansion as in progress
+	requireExpansion, err := e.startExpansion(size, replicaClients)
+	if err != nil {
+		return errors.Wrap(err, "startExpansion failed")
+	}
+	if !requireExpansion {
+		e.log.Info("No need to expand engine")
+		e.SpecSize = size
```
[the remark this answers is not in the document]
davidcheng0922: Currently, in expansion, the `e.SpecSize` will only be overridden when the expansion is completed without error.

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,114 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+		}
+
+		if !resized {
+			return fmt.Errorf("no error, but replica %s not resized", r.Name)
+		}
+
+		// If we had previously exposed the bdev, we must re-expose it after the resize.
+		if reExposeBdev {
+			nguid := commonutils.RandomID(nvmeNguidLength)
+			if err := spdkClient.StartExposeBdev(helpertypes.GetNQN(r.Name), r.Head.UUID, nguid, r.IP, strconv.Itoa(int(r.PortStart))); err != nil {
```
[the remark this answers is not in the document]
davidcheng0922: This replica will be seem as resize failed, because it doesn't expose successfully

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,114 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+			r.log.Errorf("update head failed, %v", err)
+			return err
+		}
+
+		r.log.Info("Expanding replica complete")
+	} else if bdevLvol.SpecSize > size {
+		r.log.Warnf("Found the actual size %v of replica %s is already larger than the requested size %v, will just update the spec size", bdevLvol.SpecSize, r.Name, size)
+	}
+
+	r.SpecSize = size
```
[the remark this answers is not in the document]
davidcheng0922: yeah, maybe should be `r.SpecSize = bdevLvol.SpecSize`

pkg/spdk/engine.go
```diff
@@ -1229,6 +1349,414 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+		e.log.WithFields(logrus.Fields{"failedReplicas": aggregatedErr}).
+			Error("All replicas failed to expand")
+
+		return fmt.Errorf("all replicas failed to expand; aborting RAID recreation")
+	default:
+		// partial success
+		// the failedReplica is seems as outOfSync
+		// set it to ERR
+		for replicaName := range failedReplica {
+			e.ReplicaStatusMap[replicaName].Mode = types.ModeERR
```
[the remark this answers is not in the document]
derekbit: cc @c3y1huang

pkg/spdk/engine.go
derekbit: We don't document the behavior of expansion failure in the official document.
@davidcheng0922 Can you create a DOC ticket to improve https://longhorn.io/docs/1.10.0/nodes-and-volumes/volumes/expansion/?

pkg/client/client.go
```diff
@@ -111,6 +111,22 @@ func (c *SPDKClient) ReplicaGet(name string) (*api.Replica, error) {
 	return api.ProtoReplicaToReplica(resp), nil
 }
 
+func (c *SPDKClient) ReplicaExpand(name string, size uint64) error {
+	if name == "" {
+		return fmt.Errorf("failed to delete SPDK replica: missing required parameter")
```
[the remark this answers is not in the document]
innobead: Can you help create an issue for this refactor improvement in the future? Thanks.

pkg/spdk/engine.go
```diff
@@ -1229,6 +1351,396 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+			aggregatedErr[replicaName] = err.Error()
+		}
+	}
+
+	switch {
+	case len(failedReplica) == 0:
+		// all success
+		e.log.Info("All replicas expand success")
+		return nil
+	case len(failedReplica) == len(e.ReplicaStatusMap):
```
c3y1huang, line 1662: Shouldn’t this compare `len(failedReplica)` with `len(replicaClients)` instead? Could `e.ReplicaStatusMap` include replicas that weren’t part of this expansion and give the wrong result?

pkg/spdk/engine.go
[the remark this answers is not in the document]
c3y1huang: Ah I see. Then no concurrent write happens here, you are right, no lock is needed.

pkg/client/client.go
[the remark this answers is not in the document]
davidcheng0922: https://github.com/longhorn/longhorn/issues/11708

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,119 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+	if err != nil {
+		return errors.Wrapf(err, "Get replica %s failed", r.Name)
+	}
+	bdevLvol := BdevLvolInfoToServiceLvol(&headBdevLvol)
+
+	if bdevLvol.SpecSize > size {
+		r.log.Warnf("Found the actual size %v of replica %s is already larger than the requested size %v, will just update the spec size", bdevLvol.SpecSize, r.Name, size)
+		r.SpecSize = bdevLvol.SpecSize
+		return nil
+	}
```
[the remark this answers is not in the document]
davidcheng0922: You're right, I'll remove the redundant check

pkg/client/client.go
```diff
@@ -679,6 +695,25 @@ func (c *SPDKClient) EngineWatch(ctx context.Context) (*api.EngineStream, error)
 	return api.NewEngineStream(stream), nil
 }
 
+func (c *SPDKClient) EngineExpand(ctx context.Context, name string, size uint64) error {
+	if name == "" {
+		return fmt.Errorf("failed to delete target for engine: missing required parameter")
```
derekbit, line 700:
```suggestion
		return fmt.Errorf("failed to expand engine: missing required parameter")
```

pkg/spdk/engine.go
```diff
@@ -379,6 +390,46 @@ func (e *Engine) Create(spdkClient *spdkclient.Client, replicaAddressMap map[str
...
+	// check if all replica sizes are the same
+	expectedSize := uint64(0)
+	for _, replicaSize := range replicaSizeMap {
+		if expectedSize == 0 {
+			expectedSize = replicaSize
+			continue
+		}
+
+		if expectedSize != replicaSize {
+			return fmt.Errorf("found different replica sizes: %v", replicaSizeMap)
```
derekbit, line 422:
```suggestion
			return fmt.Errorf("found different replica sizes: %+v", replicaSizeMap)
```

pkg/spdk/server.go
```diff
@@ -1095,6 +1111,30 @@ func (s *Server) EngineResume(ctx context.Context, req *spdkrpc.EngineResumeRequ
...
+		return nil, grpcstatus.Error(grpccodes.InvalidArgument, "engine name is required")
+	}
+
+	s.RLock()
+	e := s.engineMap[req.Name]
+	s.RUnlock()
+
+	if e == nil {
+		return nil, grpcstatus.Errorf(grpccodes.NotFound, "cannot find engine %v for resumption", req.Name)
+	}
```
derekbit, line 1125:
```suggestion
	if e == nil {
		return nil, grpcstatus.Errorf(grpccodes.NotFound, "cannot find engine %v for expansion", req.Name)
	}
```

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,103 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+	clusterSize, err := r.fetchClusterSize(spdkClient)
+	if err != nil {
+		return errors.Wrapf(err, "Replica %s, fetch cluster size failed", r.Name)
+
+	}
+
+	roundedSize := util.RoundUp(size, clusterSize)
+	if roundedSize != size {
+		return fmt.Errorf("replica %s rounded up spec size from %v to %v since the spec size should be multiple of MiB", r.Name, size, roundedSize)
+	}
```
derekbit, line 1061: IIRC, clusterSize is 1 MiB. longhorn-manager rounds up the size by 2MiB. The round-up might be not really necessary.

If we want to validate the size. We can move the check to longhorn-manager. https://github.com/longhorn/longhorn-manager/blob/master/webhook/resources/volume/mutator.go#L276-L280?

Can you create a ticket for the improvement?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1310,410 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	}
+	if e.SpecSize == size {
+		e.log.Infof("Engine already at requested size %v, skipping expansion", size)
+		return false, nil // no need to expand
+	}
+
+	roundedNewSize := util.RoundUp(size, helpertypes.MiB)
+	if roundedNewSize != size {
+		return false, fmt.Errorf("rounded up spec size from %v to %v since the spec size should be multiple of MiB", size, roundedNewSize)
+	}
```
derekbit, line 1397: Same question in replica.go.
Create an improvement to move the check to longhorn-manager.

pkg/spdk/engine.go
```diff
@@ -1229,6 +1310,410 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+			}
+		}()
+	}
+	if err != nil {
+		return errors.Wrap(err, "prepare raid for expansion failed")
+	}
+
+	// expand replicas
+	if err := e.expandReplicas(replicaClients, spdkClient, size); err != nil {
+		return errors.Wrap(err, "expand replica failed")
```
derekbit, line 1364:
```suggestion
		return errors.Wrap(err, "failed to expand replicas")
```

pkg/spdk/engine.go
```diff
@@ -1229,6 +1310,410 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+	}
+
+	// recreate RAID and reconnect frontend
+	if err := e.reconnectFrontend(spdkClient, bdevRaidUUID, superiorPortAllocator); err != nil {
+		return errors.Wrap(err, "reconnectFrontend failed")
```
derekbit, line 1369:
```suggestion
		return errors.Wrap(err, "failed to reconnect frontend")
```

pkg/spdk/engine.go
```diff
@@ -1229,6 +1310,410 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+				Warn("Could not verify replica expansion after failure")
+			continue
+		}
+
+		if replica.SpecSize == size {
+			e.log.WithField("replica", replicaName).
+				Info("Replica expansion succeeded despite earlier error, removing from failure list")
+
+			delete(failedReplica, replicaName)
+		}
```
shuo-wu, line 1611: What if somehow the replica lvol resize succeeds but the following nvmf expose or connect fails?

pkg/spdk/engine.go
```diff
@@ -1229,6 +1310,410 @@ func (e *Engine) validateAndUpdateReplicaNvme(replicaName string, bdev *spdktype
...
+		if err != nil {
+			return err
+		}
+
+		initiatorCreationRequired, targetCreationRequired, err := e.checkInitiatorAndTargetCreationRequirements(podIP, e.NvmeTcpFrontend.IP, e.NvmeTcpFrontend.TargetIP)
+		if err != nil {
+			return err
+		}
+
+		err = e.handleNvmeTcpFrontend(spdkClient, superiorPortAllocator, 0, targetAddress, initiatorCreationRequired, targetCreationRequired)
```
shuo-wu, line 1704: Do we need to refactor this function slightly after discarding the previous implementation `reconnectNvmeTcpFrontend`?

pkg/spdk/replica.go
davidcheng0922: https://github.com/longhorn/longhorn/issues/11718

pkg/spdk/replica.go
```diff
@@ -1043,6 +1043,103 @@ func (r *Replica) Get() (pReplica *spdkrpc.Replica) {
...
+
+	// If the bdev is exposed, we must stop exposing it before the resize.
+	reExposeBdev := false
+	if r.IsExposed {
+		if err := spdkClient.StopExposeBdev(helpertypes.GetNQN(r.Name)); err != nil && !jsonrpc.IsJSONRPCRespErrorNoSuchDevice(err) {
+			return errors.Wrapf(err, "failed to stop expose replica %v before expansion", r.Name)
+		}
+		r.IsExposed = false
+		reExposeBdev = true
+	}
```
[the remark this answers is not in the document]
davidcheng0922: It’s fine. Before expanding the replica, the bdev RAID is already deleted, so at that moment there is no engine present.

pkg/spdk/engine.go
davidcheng0922: Thank you, this is indeed a potential issue.

The extra size check is performed only when `spdkClient.BdevLvolResize` fails
Failures from `StartExposeBdev` and `connectNVMfBdev` are considered as replica expansion failures.

pkg/spdk/engine.go
davidcheng0922: I refactored `handleNvmeTcpFrontend` to clean up the logic without modifying how the new frontend is configured. I introduced two helper functions:

`isNvmeFrontendConfigured` – to check if the frontend has already been configured.

`configureNvmeTcpFrontend` – extracted from the original setup logic to make the main function cleaner and easier to read.

derekbit (approved):
LGTM

derekbit:
@davidcheng0922 Thank you. You can update longhorn-instance-manager PR.

PR #394, opened 2025-08-20
fix(backup): fallback to full backup if comparing snapshot is not an ancestor

c3y1huang:

Issue longhorn/longhorn#11461

Previously, attempting to back up an older snapshot after a new snapshot was backed up would fail with `invalid snapshot lvol range` because the comparing snapshot was not a valid ancestor.

With this change, the backup falls back to a full backup when the comparing snapshot is not an ancestor, ensuring that older snapshots can be backed up reliably.

`None`

`None`


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3874/console

PR #403, opened 2025-09-05
revert: feat(engine): add I/O suspension during replica deletion

hookak:
This reverts commit f640e455c81df7c2d7b008263beae9528ace1a2f. Longhorn 11109

Issue # https://github.com/longhorn/longhorn/issues/11109

This PR reverts commit `f640e455c81df7c2d7b008263beae9528ace1a2f`.(https://github.com/longhorn/longhorn-spdk-engine/pull/376), which suspended I/O during replica deletion.
Although that change did not cause functional failures, it can negatively affect client I/O while a replica is being deleted. SPDK already performs an internal quiesce(suspension of I/O) during base bdev removal, so keeping the additional suspend increases client-visible impact.


derekbit (approved):
LGTM

PR #410, opened 2025-09-16
fix: avoid invalid state update during replica deletion

shuo-wu:
Longhorn 10112

Issue https://github.com/longhorn/longhorn/issues/10112


shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/4135/

PR #411, opened 2025-09-19
fix: improve engine and replica logs

shuo-wu:

This helps with the snapshot/rebuilding-related issue debugging


derekbit (approved):
LGTM

PR #468, opened 2026-01-22
fix(grpc): disable grpc service discovery to prevent dns flooding

COLDTURNIP:

Issue longhorn/longhorn#12521


derekbit (approved):
LGTM

PR #471, opened 2026-02-05
feat(backupRestore): move backup restore workflow from replica to engine level

davidcheng0922:

https://github.com/longhorn/longhorn/issues/9277

LEP: https://github.com/longhorn/longhorn/pull/12609

Preserve the existing `replica.BackupRestore` implementation for future use.

e2e test pass:
- negativeANDbackup
- negativeANDdr-volume
- regressionANDbackup

https://ci.longhorn.io/job/private/job/longhorn-e2e-test/6820/
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/6821/
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/6829/

### DIFF
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -123,7 +123,6 @@ type Engine struct {
 	SpecSize   uint64
 	ActualSize uint64
 	Frontend   string
-	Endpoint   string
 
 	ctrlrLossTimeout     int
 	fastIOFailTimeoutSec int
@@ -717,7 +716,6 @@ func (e *Engine) getWithoutLock() (res *spdkrpc.Engine) {
 		ReplicaModeMap:        map[string]spdkrpc.ReplicaMode{},
 		Snapshots:             map[string]*spdkrpc.Lvol{},
 		Frontend:              e.Frontend,
-		Endpoint:              e.Endpoint,
 		State:                 string(e.State),
 		ErrorMsg:              e.ErrorMsg,
 		IsExpanding:           e.isExpanding,
@@ -2117,7 +2115,6 @@ func (e *Engine) BackupRestore(spdkClient *spdkclient.Client, backupUrl, endpoin
 	defer func() {
 		if err != nil {
 			e.IsRestoring = false
-			e.Endpoint = ""
 		}
 	}()
 
@@ -2289,9 +2286,6 @@ func (e *Engine) completeBackupRestore(spdkClient *spdkclient.Client, backupSnap
 	if e.restore != nil {
 		oldSnapshotName = e.restore.SnapshotName
 	}
-
-	// Clear the endpoint; the EngineFrontend will tear down the initiator after doneCh closes.
-	e.Endpoint = ""
 
 	// Stop the temporary NVMe-TCP target and release its port.
 	if e.Frontend == types.FrontendEmpty {
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -1,6 +1,7 @@
 package spdk
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"strconv"
@@ -21,6 +22,8 @@ import (
 	"github.com/longhorn/go-spdk-helper/pkg/jsonrpc"
 	"github.com/longhorn/types/pkg/generated/spdkrpc"
 
+	btypes "github.com/longhorn/backupstore/types"
+	butil "github.com/longhorn/backupstore/util"
 	commonbitmap "github.com/longhorn/go-common-libs/bitmap"
 	commonnet "github.com/longhorn/go-common-libs/net"
 	commonutils "github.com/longhorn/go-common-libs/utils"
@@ -110,6 +113,11 @@ func (m *MockReplicaAdder) ReplicaAddFinish(srcReplicaServiceCli, dstReplicaServ
 type Engine struct {
 	sync.RWMutex
 
+	ctx       context.Context
+	cancelCtx context.CancelFunc
+
+	restore *EngineRestore
+
 	Name       string
 	VolumeName string
 	SpecSize   uint64
@@ -179,7 +187,12 @@ func NewEngine(engineName, volumeName, frontend string, specSize uint64, engineU
 	}
 	log.WithField("specSize", roundedSpecSize)
 
+	ctx, cancelCtx := context.WithCancel(context.Background())
+
 	e := &Engine{
+		ctx:       ctx,
+		cancelCtx: cancelCtx,
+
 		Name:       engineName,
 		VolumeName: volumeName,
 		Frontend:   frontend,
@@ -592,6 +605,11 @@ func (e *Engine) Delete(spdkClient *spdkclient.Client, superiorPortAllocator *co
 	}()
 
 	e.log.Info("Deleting engine")
+	if e.IsRestoring && e.restore != nil {
+		e.log.Info("Canceling volume restoration before engine deletion")
+		e.cancelCtx()
+		e.restore.Stop()
+	}
 
 	e.log.Infof("Stopping to expose RAID bdev for engine %s", e.Name)
 	switch e.Frontend {
@@ -782,6 +800,9 @@ func (e *Engine) ReplicaAdd(spdkClient *spdkclient.Client, dstReplicaName, dstRe
 	// Syncing with the SPDK TGT server only when the engine is running.
 	if e.State != types.InstanceStateRunning {
 		return fmt.Errorf("invalid state %v for engine %s replica %s add start", e.State, e.Name, dstReplicaName)
+	}
+	if e.IsRestoring {
+		return fmt.Errorf("cannot add replica %s while engine %s restore is in progress", dstReplicaName, e.Name)
 	}
 
 	if _, exists := e.ReplicaStatusMap[dstReplicaName]; exists {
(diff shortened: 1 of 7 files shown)


pkg/spdk/engine.go
```diff
@@ -2850,202 +2871,299 @@ func (e *Engine) BackupStatus(backupName, replicaAddress string) (*spdkrpc.Backu
...
-			if err := e.completeBackupRestore(spdkClient); err != nil {
-				e.log.WithError(err).Warn("Failed to complete backup restore")
-			}
-		}()
+	// Only wait backup restore if backup restore initiation is successful
+	go func() {
+		if err := e.completeBackupRestore(spdkClient, superiorPortAllocator); err != nil {
+			e.log.WithError(err).Warn("Failed to complete backup restore")
+		}
 	}()
```
shuo-wu: It's better to put this immediately after the frontend handling. Then the cleanup will be executed even if the restore fails

pkg/spdk/engine.go
```diff
@@ -2850,202 +2871,299 @@ func (e *Engine) BackupStatus(backupName, replicaAddress string) (*spdkrpc.Backu
...
+				e.log.WithError(err).Errorf("Failed to delete existing snapshot %v of the restored volume", oldSnapshotName)
+				return errors.Wrapf(err, "failed to delete snapshot %v of the restored volume", oldSnapshotName)
+			}
+		}
+	}
+
+	newSnapshotName := util.UUID()
+	if _, err := e.SnapshotCreate(spdkClient, newSnapshotName); err != nil {
+		return errors.Wrapf(err, "failed to create snapshot %v after restore", newSnapshotName)
+	}
```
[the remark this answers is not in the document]
davidcheng0922: I agree. For now, I think we can keep the previous design. 
I'll use the backup snapshot name when creating the new snapshot, and fall back to a UUID if it is empty.

pkg/spdk/engine.go
```diff
@@ -2850,201 +2871,316 @@ func (e *Engine) BackupStatus(backupName, replicaAddress string) (*spdkrpc.Backu
...
+	e.log.Infof("Successfully created restored snapshot %v", e.restore.SnapshotName)
+
+	return nil
+}
+
+func (e *Engine) waitForRestoreComplete() error {
+	e.log.Info("Waiting for restore to complete")
+
+	periodicChecker := time.NewTicker(time.Duration(restorePeriodicRefreshInterval.Seconds()) * time.Second)
+	defer periodicChecker.Stop()
```
shuo-wu: NIT: @derekbit introduced dependency `"github.com/avast/retry-go/v4"` for this kind of retry logic. e.g., https://github.com/longhorn/go-spdk-helper/pull/265

derekbit:
Hello @davidcheng0922 
You can continue the development. The engine separation PR is merged. Thanks.

pkg/spdk/engine.go
```diff
@@ -1918,211 +1961,339 @@ func (e *Engine) BackupStatus(backupName, replicaAddress string) (*spdkrpc.Backu
...
+	}()
+
+	resp = &spdkrpc.EngineBackupRestoreResponse{
 		Errors: map[string]string{},
 	}
 
+	if len(e.ReplicaStatusMap) == 0 {
+		return resp, nil, fmt.Errorf("cannot restore backup %s: no replicas available", backupUrl)
+	}
+
```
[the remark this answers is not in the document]
davidcheng0922: Fix

derekbit:
@davidcheng0922 Ready for review?

pkg/spdk/engine.go
```diff
@@ -18,6 +19,8 @@ import (
 	grpcstatus "google.golang.org/grpc/status"
 
 	"github.com/longhorn/backupstore"
+	btypes "github.com/longhorn/backupstore/types"
+	butil "github.com/longhorn/backupstore/util"
 	"github.com/longhorn/go-spdk-helper/pkg/jsonrpc"
```
derekbit, line 22: Reorganize the imports

pkg/spdk/server_engine.go
```diff
@@ -507,22 +507,64 @@ func (s *Server) EngineBackupStatus(ctx context.Context, req *spdkrpc.BackupStat
...
-
-	if e == nil {
+	e, exist := s.engineMap[req.EngineName]
+	if !exist {
+		s.RUnlock()
 		return nil, grpcstatus.Errorf(grpccodes.NotFound, "cannot find engine %v for restoring backup", req.EngineName)
 	}
 
-	return e.BackupRestore(spdkClient, req.BackupUrl, req.EngineName, req.SnapshotName, req.Credential, req.ConcurrentLimit)
+	// Backup restore is only supported for DR engines, which have no associated
```
derekbit:
```
	// Backup restore is only supported for DR engines, which have no associated
```

A bit confused. Why does `Backup restore` only support DR engine?

davidcheng0922:
> @davidcheng0922 Ready for review?

Yes, I rebase with the latest version. Thank you

pkg/spdk/enginefrontend.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"go.uber.org/multierr"
 
+	commonbitmap "github.com/longhorn/go-common-libs/bitmap"
 	"github.com/longhorn/longhorn-spdk-engine/pkg/client"
 	"github.com/longhorn/longhorn-spdk-engine/pkg/types"
```
derekbit: Reorg the imports

pkg/spdk/server_engine.go
```diff
@@ -540,22 +540,64 @@ func (s *Server) EngineBackupStatus(ctx context.Context, req *spdkrpc.BackupStat
...
+		frontend.RLock()
+		engineName := frontend.EngineName
+		frontendType := frontend.Frontend
+		frontendName := frontend.Name
+		frontend.RUnlock()
+
+		if engineName == req.EngineName && frontendType != types.FrontendEmpty {
+			s.RUnlock()
+			return nil, grpcstatus.Errorf(grpccodes.FailedPrecondition, "cannot restore backup: engine %v has a non-empty frontend %v", req.EngineName, frontendName)
+		}
```
derekbit, line 567: Okay, actually, an enginefrontend for FrontendEmpty frontend exists. We may update the above comment to avoid confusion.

pkg/spdk/server_engine.go
```diff
@@ -540,22 +540,64 @@ func (s *Server) EngineBackupStatus(ctx context.Context, req *spdkrpc.BackupStat
...
+	tempEF := NewEngineFrontend(
+		e.Name+"-restore",
+		e.Name,
+		volumeName,
+		types.FrontendSPDKTCPBlockdev,
+		specSize,
+		types.DefaultUblkQueueDepth,
+		types.DefaultUblkNumberOfQueue,
+		throwawayUpdateCh,
+	)
```
derekbit, line 599: I'm thinking if we can put it in a map similar to EngineFrontendMap to improve the observability and monitoring.

pkg/spdk/engine.go
```diff
@@ -2050,212 +2071,341 @@ func (e *Engine) BackupStatus(backupName, replicaAddress string) (*spdkrpc.Backu
...
 	}()
 
-	for replicaName, replicaStatus := range e.ReplicaStatusMap {
-		e.log.Infof("Restoring backup on replica %s address %s", replicaName, replicaStatus.Address)
+	// The frontend (NVMe-TCP initiator) is managed by EngineFrontend.
+	// Store the provided endpoint so EngineRestore.OpenVolumeDev can access the block device.
+	// Also copy it into e.restore.endpoint so OpenVolumeDev does not need to re-acquire the
+	// engine lock (which would deadlock — BackupRestore already holds e.Lock()).
+	e.log.Infof("Using endpoint %v for backup restore", endpoint)
+	e.Endpoint = endpoint
```
derekbit: Do we need to keep `e.Endpoint`?

pkg/spdk/enginefrontend.go
```diff
@@ -2662,3 +2663,113 @@ func (ef *EngineFrontend) RecoverFromHost(spdkClient *spdkclient.Client) error {
...
+	}
+
+	engine.RLock()
+	targetAddress := net.JoinHostPort(engine.NvmeTcpTarget.IP, strconv.Itoa(int(engine.NvmeTcpTarget.Port)))
+	engine.RUnlock()
+
+	ef.log.Infof("Setting up NVMe-TCP initiator for backup restore, target: %s", targetAddress)
+	if err := ef.createNvmeTcpFrontend(spdkClient, targetAddress); err != nil {
+		return nil, errors.Wrapf(err, "failed to setup NVMe-TCP frontend for backup restore")
+	}
```
derekbit, line 2725: Do we need to clean up Nvme-TCP Target for restore if it is failed?

pkg/spdk/engine.go
```diff
@@ -2050,212 +2071,341 @@ func (e *Engine) BackupStatus(backupName, replicaAddress string) (*spdkrpc.Backu
...
-	e.log.Infof("Creating raid bdev %s with replicas %+v before finishing restoration", e.Name, replicaBdevList)
-	if _, err := spdkClient.BdevRaidCreate(e.Name, spdktypes.BdevRaidLevel1, 0, replicaBdevList, ""); err != nil {
-		if !jsonrpc.IsJSONRPCRespErrorFileExists(err) {
-			e.log.WithError(err).Errorf("Failed to create raid bdev before finishing restoration")
-			return err
-		}
+	e.log.Infof("Creating snapshot %v for the restored volume", newSnapshotName)
+	if _, createErr := e.SnapshotCreate(spdkClient, newSnapshotName); createErr != nil {
+		e.log.WithError(createErr).Warnf("Failed to create snapshot %v for the restored volume", newSnapshotName)
 	}
```
derekbit, line 2363: One question, do we really need to create an new snapshot?
cc @shuo-wu

pkg/spdk/engine.go
```diff
@@ -1918,211 +1961,339 @@ func (e *Engine) BackupStatus(backupName, replicaAddress string) (*spdkrpc.Backu
...
+	}
+
 	backupInfo, err := backupstore.InspectBackup(backupUrl)
 	if err != nil {
-		for _, replicaStatus := range e.ReplicaStatusMap {
-			resp.Errors[replicaStatus.Address] = err.Error()
-		}
-		return resp, nil
+		return resp, nil, err
 	}
```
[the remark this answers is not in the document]
derekbit: The concerns seems valid. WDYT? @davidcheng0922

pkg/spdk/engine.go
davidcheng0922: Addressed, thank you

pkg/spdk/server_engine.go
davidcheng0922: Addressed, thank you. I update the comment

pkg/spdk/enginefrontend.go
davidcheng0922: Addressed, thank you

pkg/spdk/server_engine.go
davidcheng0922: Addressed, thank you

pkg/spdk/engine.go
davidcheng0922: Addressed, thank you.

pkg/spdk/enginefrontend.go
davidcheng0922: Add the cleanup function, thank you

pkg/spdk/server_engine.go
davidcheng0922: I think keeping a map just for the temporary restore frontend is probably not worth the cost. We already have restore status for monitoring, and I added more logs in the BackupRestore path to improve observability. WDYT?

pkg/spdk/engine.go
davidcheng0922: Not really, the snapshots are not used at the moment.

Based on our previous discussion, they are intended for a future improvement—to reuse snapshots in backup restore failure scenarios. I will create a follow-up ticket for this enhancement after this tickets are merged.

For now, I’ve kept the snapshot operations as they were. I can also remove them and reintroduce this logic when we implement the failure recovery (rescue) mechanism.

pkg/spdk/engine.go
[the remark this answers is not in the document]
davidcheng0922: Addressed, thank you

pkg/spdk/server_engine.go
derekbit: Sounds good. 
Another question, is it possible that the restore is stuck forever?

davidcheng0922: Normally, if there is a networking issue or the backup target fails, we should get a restore error, so the restore should not hang forever. 

The extreme corner case is when the restore is effectively stuck or failed, but no error is returned. In that case, it can hang. I'm not sure if this happens, but if it does hang, deleting the volume will cancel the restore and clean up the resources. 

We do not currently have a restore timeout because it is difficult to define a reasonable value, and restore duration is unpredictable.

pkg/spdk/engine.go
derekbit: Can we remove `Endpoint`? It seems useless.

davidcheng0922: checking

davidcheng0922: I think it is ok to remove it. thank you

derekbit (approved):
LGTM

PR #494, opened 2026-04-12
feat(switchover): implement NVMe/TCP multipath live switchover

derekbit:

Longhorn/longhorn#7124
Longhorn/longhorn#10469


pkg/spdk/enginefrontend.go
```diff
@@ -1183,15 +1771,391 @@ func (ef *EngineFrontend) SwitchOverTarget(spdkClient *spdkclient.Client, newEng
...
+		if err := ef.setRemoteEngineTargetANAState(newTargetIP, newEngineName, NvmeTCPANAStateInaccessible); err != nil {
+			setErr := errors.Wrapf(ErrSwitchOverTargetInternal,
+				"failed to set new engine target %s ANA to inaccessible before connect: %v",
+				newEngineName, err)
+			ef.Lock()
+			ef.ErrorMsg = setErr.Error()
+			ef.Unlock()
+			*updateRequired = true
+			return setErr
+		}
```
davidcheng0922: Do we need a best-effort rollback defer for handling after step 1 success when steps 2, 3 or 4 fail?

pkg/spdk/enginefrontend.go
```diff
@@ -30,18 +31,21 @@ import (
...
+	VolumeNQN   string
+	VolumeNGUID string
+	SpecSize    uint64
+	ActualSize  uint64
 
 	Frontend string
 	Endpoint string
 
-	EngineIP string
-
```
davidcheng0922, line 43: Remove EngineIP, when engine frontend is empty, is it still work? 

Such like in `Expand()`

```go
	var engineServiceAddress string
	if ef.NvmeTcpFrontend != nil {
		targetAddress = net.JoinHostPort(ef.NvmeTcpFrontend.TargetIP, strconv.Itoa(int(ef.NvmeTcpFrontend.TargetPort)))
		engineServiceAddress = net.JoinHostPort(ef.NvmeTcpFrontend.TargetIP, strconv.Itoa(types.SPDKServicePort))
	}

	engineSpdkClient, err := GetServiceClient(engineServiceAddress)
	if err != nil {
		ef.Unlock()
		return errors.Wrapf(err, "failed to get SPDK client to expand engine frontend %v", ef.Name)
	}
```

It cannot be expanded when the frontend is empty (DR volume)

davidcheng0922: Also in `snapshotOperation`

pkg/spdk/enginefrontend.go
derekbit: No, a rollback defer is not needed here.

- Step 1 sets the new target to inaccessible. This is inherently safe. An `inaccessible` path receives no I/O. If steps 2, 3, or 4 fail:

- Step 2 (connect) fails: The new path is `inaccessible` and either not connected or already-connected. No I/O is routed to it. The old path is still `optimized` and fully functional. 

- Step 3 (wait controller live) fails: The new path is connected but `inaccessible`. The kernel won't route I/O through it.

- Step 4 (set `non-optimized`) fails: The new path stays `inaccessible`. The old path is still `optimized`. The control plane can retry the entire preparing phase idempotently.

WDYT?

pkg/spdk/enginefrontend.go
derekbit: Good catch! I thought EnigneIP is redundant and forgot the DR volume case.
Fixed. Thank you.

derekbit:
@davidcheng0922 Ready for review. Thank you.

pkg/spdk/enginefrontend.go
davidcheng0922: Make sense

pkg/spdk/engine.go
```diff
@@ -308,26 +311,125 @@ func (e *Engine) createNVMeTCPTarget(spdkClient *spdkclient.Client, superiorPort
...
+	e.RLock()
+	if e.NvmeTcpTarget == nil {
+		e.RUnlock()
+		return fmt.Errorf("engine %s does not have an NVMe/TCP target", e.Name)
+	}
+	nqn := e.NvmeTcpTarget.Nqn
+	ip := e.NvmeTcpTarget.IP
+	port := e.NvmeTcpTarget.Port
+	e.RUnlock()
+
```
davidcheng0922, line 411: Do we lock for the entire function?

T1: SetANAState reads nqn/ip/port under read lock
T2: Releases lock, starts SPDK RPC
T3: Expand acquires write lock, deletes subsystem
T4: Expand recreates subsystem with currentANAState (optimized)
T5: SetANAState RPC fails (subsystem not found) OR succeeds on wrong subsystem
T6: SetANAState updates e.NvmeTcpTarget.ANAState to inaccessible

pkg/spdk/enginefrontend.go
```diff
@@ -160,6 +202,355 @@ func NewEngineFrontend(engineFrontendName, engineName, volumeName, frontend stri
...
+		}
+	}
+
+	syncFn := ef.syncRemoteEngineTargetANAStates
+	if ef.syncRemoteEngineTargetANAStatesFn != nil {
+		syncFn = ef.syncRemoteEngineTargetANAStatesFn
+	}
+
+	const maxAttempts = 5
+	const retryInterval = 200 * time.Millisecond
```
davidcheng0922: to global?

pkg/spdk/engine.go
davidcheng0922: I miss click the part, I mean the in function  `SetTargetListenerANAState()`

I think it will be safer to hold the lock in the entire function

derekbit: Good catch! Add a lock to protect the all operations. Thanks.

pkg/spdk/enginefrontend.go
derekbit: Add global constants. Thanks.

pkg/spdk/enginefrontend.go
```diff
@@ -160,6 +211,352 @@ func NewEngineFrontend(engineFrontendName, engineName, volumeName, frontend stri
...
+			"targetIP":      targetIP,
+			"targetPort":    targetPort,
+		}).Warn("Failed to sync remote target ANA state during switchover, retrying")
+
+		select {
+		case <-time.After(anaSyncRetryInterval):
+		case <-ef.stopCh:
+			return syncErr
+		}
+	}
```
davidcheng0922, line 551: Should we keep the same pattern as `retry.Do`?

derekbit: Using a generic `retry.Do` style here looks cleaner, but this code depends on two specific behaviors:

- If `stopCh` is closed, it returns the last `syncErr` instead of a cancellation error.
- The retrying log is emitted only when there will actually be another retry, not on the final failure.

So if we want the same semantics, the current loop is clearer and safer.

davidcheng0922: Got it

davidcheng0922 (dismissed):
LGTM. Just make sure DR volumes are considered.

davidcheng0922 (approved):

PR #500, opened 2026-04-30
refactor: v2 linked clone 

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/12552

We introduced a special kind of empty snapshot lvol called `entrypoint lvol` for src replicas. Then, all clone replicas use an entrypoint lvol as their `ActiveChain[0]`, which is like a backing image. After that, clone replicas should be both readable and writable like normal replicas.

Notice Claude Opus 4.6 generates the codes based on the following prompt/plan. And the plan is mainly for simple reference only. Because there are several rounds of improvement after that. 

```
First, refactor the existing linked-clone feature for repo `longhorn-spdk-engine` based on the descriptions below:
- The source replica:
  1. will prepare an entrypoint lvol (which is actually an empty snapshot lvol) from the requested snapshot
  2. record all entrypoints and all clone replicas pointing to it. Even for a src replica detected from an existing disk after restart, it should recover these info as well.
  3. avoid or ignore all entrypoints when checking or constructing its snapshot map 
  4. automatically clean up the entrypoint when there is no clone replica pointing to it
  5. forbid the replica deletion when cleanup is true, while there are still clone replicas pointing to it
  6. forbid the snapshot deletion when there is still a valid entrypoint pointing to it
  7. sync func should verify (and probably repair) all entrypoint lvols for the src replica, and linked-clone source information for clone replicas. 
- The clone replica itself:
  1. grows a new head lvol from the entrypoint lvol.
  2. record its entrypoint and the source replica. Even for the replica detected from an existing disk after restart, it should recover these info as well.
  3. sync func should verify linked-clone source information for clone replicas. And for case that the clone replica's root parent somehow becomes `the source snapshot directly`, the sync func should repair it.

```


derekbit:
cc @davidcheng0922 @COLDTURNIP for review

pkg/spdk/replica.go
```diff
@@ -2224,6 +2512,34 @@ func (r *Replica) SnapshotCloneDstStatusCheck() (status *spdkrpc.ReplicaSnapshot
 
 func (r *Replica) SnapshotCloneDstFinish(spdkClient *spdkclient.Client, cloneMode spdkrpc.CloneMode) (err error) {
 	if cloneMode == spdkrpc.CloneMode_CLONE_MODE_LINKED_CLONE {
+		if r.Head == nil {
+			return fmt.Errorf("cannot find the head for replica %s linked-clone finish", r.Name)
+		}
+		if r.cloneEntrypointLvolName == "" {
+			return fmt.Errorf("clone entrypoint lvol name is empty for replica %s linked-clone finish", r.Name)
+		}
+		epAlias := spdktypes.GetLvolAlias(r.LvsName, r.cloneEntrypointLvolName)
```
davidcheng0922, line 2565: Question,  the clone entrypoint is created on the source replica's lvstore, but here we resolve it using the destination replica's LvsName. If src and dst are on different disks/LVSes, will it be a issue?

pkg/spdk/replica.go
```diff
@@ -2354,53 +2670,183 @@ func (r *Replica) snapshotLinkedCloneSrcStart(spdkClient *spdkclient.Client, sna
...
+		// Auto-cleanup: no clone replicas left
+		if len(epInfo.CloneReplicas) == 0 {
+			r.log.Infof("Clone entrypoint %s has no remaining clone replicas, cleaning up", epLvolName)
+			epAlias := spdktypes.GetLvolAlias(r.LvsName, epLvolName)
+			if _, err := spdkClient.BdevLvolDelete(epAlias); err != nil && !jsonrpc.IsJSONRPCRespErrorNoSuchDevice(err) {
+				r.log.WithError(err).Errorf("Failed to delete unused clone entrypoint %s", epLvolName)
+				continue
+			}
+			delete(r.cloneEntrypointMap, epLvolName)
+		}
```
davidcheng0922, line 2887: Is it possible to hit a race like this?

t1: ReplicaSnapshotCloneSrcStart creates the entrypoint
t2: syncCloneEntrypoints runs
t3: SnapshotCloneDstFinish reparents the dst replica


and syncCloneEntrypoints not yet to find the children and delete the entrypoint?

pkg/spdk/engine.go
```diff
@@ -1931,6 +1950,91 @@ func (e *Engine) SnapshotClone(snapshotName, srcEngineName, srcEngineAddress str
...
+				}
+			}()
+
+			if err := cli.ReplicaSnapshotCloneDstStart(dst.name, snapshotName, dst.srcReplicaName, dst.srcReplicaAddr, spdkrpc.CloneMode_CLONE_MODE_LINKED_CLONE); err != nil {
+				mu.Lock()
+				if firstErr == nil {
+					firstErr = fmt.Errorf("replica %s: SnapshotCloneDstStart failed: %w", dst.name, err)
+				}
+				mu.Unlock()
+			}
```
davidcheng0922, line 2068: If ReplicaSnapshotCloneDstStart() fails here, do we also need to update the corresponding replica state, for example mark it as ERR?

pkg/spdk/replica.go
```diff
@@ -457,16 +494,202 @@ func (r *Replica) construct(bdevLvolMap map[string]*spdktypes.BdevInfo) (err err
...
+// recoverCloneReplicaInfo detects if this replica is a linked clone by checking
+// whether ActiveChain[0] is a clone entrypoint lvol.
+func (r *Replica) recoverCloneReplicaInfo() {
+	r.isCloneReplica = false
+	r.cloneSourceReplicaName = ""
+	r.cloneEntrypointLvolName = ""
+
+	if len(r.ActiveChain) == 0 || r.ActiveChain[0] == nil || !IsCloneEntrypointLvol(r.ActiveChain[0].Name) {
+		return
+	}
```
davidcheng0922, line 569: I'm not sure if my understanding the restart-repair flow correctly:

1. syncCloneReplicaInfo() wants to repair the case where the clone root parent points to the source snapshot directly instead of the entrypoint.

2. After restart, constructActiveChainFromSnapshotLvolMap() leaves the chain base nil if the entrypoint lvol is already missing. 

3. recoverCloneReplicaInfo() only recognizes a linked-clone replica when ActiveChain[0] is still the entrypoint.

4. That means isCloneReplica is no longer recovered in this case, and 
syncCloneReplicaInfo() returns early before the repair logic runs.

If the entrypoint is already gone at restart time, will this state still be recoverable, or will the clone metadata be lost permanently?

pkg/spdk/engine.go
```diff
@@ -1803,51 +1829,21 @@ type replicaCandidate struct {
 	address string
 }
 
-func (e *Engine) SnapshotClone(snapshotName, srcEngineName, srcEngineAddress string, cloneMode spdkrpc.CloneMode) (err error) {
+func (e *Engine) SnapshotClone(snapshotName, srcEngineName, srcEngineAddress string, cloneMode spdkrpc.CloneMode, dstReplicaSrcReplicaPairMap map[string]string) (err error) {
```
derekbit, line 1869: Should we reject clone request if replicas are being rebuilt?

pkg/spdk/engine.go
```diff
@@ -1886,6 +1882,49 @@ func (e *Engine) SnapshotClone(snapshotName, srcEngineName, srcEngineAddress str
...
 	}
 
+	if cloneMode == spdkrpc.CloneMode_CLONE_MODE_LINKED_CLONE {
+		if len(dstReplicaSrcReplicaPairMap) > 0 {
+			// New path: manager provided an explicit dst→src replica name map (proxy API >= 7).
+			return e.snapshotCloneLinkedN(snapshotName, dstReplicaSrcReplicaPairMap, srcReplicaCandidates)
+		}
+		// Backward-compat: map not provided (older IM). Fall through to the single-replica
+		// deep-copy path so the operation still completes, just without N-replica parallelism.
+		e.log.Warnf("SnapshotClone: DstReplicaSrcReplicaPairMap is empty for linked-clone; falling back to 1-replica path")
```
derekbit: For the upgrade from v1.11.x to v1.12.x, users need to restart the IM pods, so the handling of the backward compatibility is not required. WDYT?

pkg/spdk/engine.go
```diff
@@ -1915,6 +1950,91 @@ func (e *Engine) SnapshotClone(snapshotName, srcEngineName, srcEngineAddress str
 	return dstReplicaServiceCli.ReplicaSnapshotCloneDstStart(dstReplicaName, snapshotName, srcReplicaName, srcReplicaAddress, cloneMode)
 }
 
+// snapshotCloneLinkedN performs SnapshotCloneDstStart on ALL RW dst replicas simultaneously
+// (linked-clone mode only). It uses the explicit dst→src replica name map provided by the
+// manager (which already ran the scheduler) instead of auto-detecting by IP+lvsUUID co-location.
+// Must be called with the engine lock held.
```
derekbit, line 1991: Why don't we use the auto-detected IP+lvsUUID?

pkg/spdk/engine.go
```diff
@@ -1915,6 +1950,91 @@ func (e *Engine) SnapshotClone(snapshotName, srcEngineName, srcEngineAddress str
...
+			e.log.Warnf("Dst replica %s is not RW (mode=%v), skipping", dstName, dstStatus.Mode)
+			continue
+		}
+
+		srcCand, ok := srcReplicaCandidates[srcName]
+		if !ok {
+			e.log.Errorf("Src replica %s not found in src engine replica list, marking dst replica %s ERR", srcName, dstName)
+			dstStatus.Mode = types.ModeERR
+			continue
+		}
```
derekbit, line 2024: Do we need to mark the replica as ERR if it is in WO state?

pkg/spdk/engine.go
```diff
@@ -1915,6 +1950,91 @@ func (e *Engine) SnapshotClone(snapshotName, srcEngineName, srcEngineAddress str
...
+				if errClose := cli.Close(); errClose != nil {
+					e.log.WithError(errClose).Warnf("Failed to close client for dst replica %s", dst.name)
+				}
+			}()
+
+			if err := cli.ReplicaSnapshotCloneDstStart(dst.name, snapshotName, dst.srcReplicaName, dst.srcReplicaAddr, spdkrpc.CloneMode_CLONE_MODE_LINKED_CLONE); err != nil {
+				mu.Lock()
+				if firstErr == nil {
+					firstErr = fmt.Errorf("replica %s: SnapshotCloneDstStart failed: %w", dst.name, err)
+				}
```
derekbit: Can we aggregate the errors?

derekbit (commented):
One question: should we rename linked-clone to fast-clone?

pkg/spdk/replica.go
```diff
@@ -2354,53 +2670,183 @@ func (r *Replica) snapshotLinkedCloneSrcStart(spdkClient *spdkclient.Client, sna
...
+			return err
 		}
+		epInfo = r.cloneEntrypointMap[epLvolName]
 	}
 
-	snapLvolName := GetReplicaSnapshotLvolName(r.Name, snapshotName)
-	snapLvol := r.SnapshotLvolMap[snapLvolName]
-	if snapLvol == nil {
-		return fmt.Errorf("cannot find snapshot %s for src replica %s", snapshotName, r.Name)
+	epInfo.CloneReplicas[dstReplicaName] = true
```
derekbit, line 2755: When will the replica be deleted from the `CloneReplicas` map? After the cloned replica is deleted?

pkg/spdk/replica.go
shuo-wu: For now, we always assume that src and dst replicas are on the same disk. 
In the future, if we allow cross-disk link clone, the caller (which should be `SnapshotCloneDstStart` then `longhorn-manager`) should tell the src replica lvs name.

pkg/spdk/replica.go
shuo-wu: Yes, it's possible. I already added an extra check (`r.snapshotCloningSrcCache`) for this. But I am not sure if this race is fully eliminated.

pkg/spdk/replica.go
shuo-wu: In func `syncCloneEntrypoints` line 2854, we will directly reconstruct all ep lvol caches for src replica `r.cloneEntrypointMap`

pkg/spdk/engine.go
shuo-wu: Typically, this link clone should be triggered for brand-new volumes only, hence all replicas should be RW. Right now, I have not added such an explicit check and just mark unexpected replicas as ERR.

pkg/spdk/engine.go
shuo-wu: YES. Removed the Backward part and allow new path for linked clone only

pkg/spdk/engine.go
shuo-wu: Do you mean the existing detection implementation? That part is for picking up the src replica when finding a matching IP+lvsUUID. In other words, if there are multiple src candidates on the same disk, this pick may choose a src replica other than what longhorn-manager scheduler selects.

pkg/spdk/engine.go
shuo-wu: YES. Because the upper longhorn-manager already determined that this dst replica depends on the selected src replica. If the src replica is unavailable, we should create a new dst replica and pick up another healthy src replica, rather than trying to rescue here.

pkg/spdk/engine.go
shuo-wu: Yes. Updated.

pkg/spdk/engine.go
shuo-wu: Yes. Updated.

pkg/spdk/replica.go
shuo-wu: I see. Do you mean the case below?
1. A clone replica somehow lost its ep lvol, and its root snap lvol was directly pointing to the src replica snap lvol
2. A restart happened, and the spdk server tried to reconstruct the replica cache
3. The new replica cache executed `construct` first. And in `construct()`, `recoverCloneReplicaInfo` considered this replica as non-clone replica, because the parant of the root snap is src replica snap rather than ep lvol then `r.ActiveChain[0]` is nil
4. The new replica cache then executed `syncCloneReplicaInfo`, which returned in the beginning due to the check `if !r.isCloneReplica {`, no chance to reach the repair logic at all!


This case is a little bit complicated, and your concern is fair. 
I am not sure whether it's really worth repairing the clone replica for these cases. Do you think it's better to blindly mark all kinds of unexpected dst replicas as ERR?

cc @derekbit

derekbit (commented):
LGTM.

The PR longhorn/types is not submitted.

pkg/spdk/replica.go
davidcheng0922: Fair, maybe it is too little chance to happen, maybe not worth to fix now

pkg/spdk/replica.go
davidcheng0922: ok, thank you

pkg/spdk/replica.go
derekbit: We can create an improvement ticket for future improvement.

derekbit:
Hello @shuo-wu 
longhorn/types PR is merged.

shuo-wu:
I just added a summary in the longhorn-manager PR https://github.com/longhorn/longhorn-manager/pull/4732. I think that is pretty helpful for the whole feature's review. If you are interested, PTAL!

derekbit:
@davidcheng0922 Do you have any concern? If no, it looks good to merge.

davidcheng0922:
> @davidcheng0922 Do you have any concern? If no, it looks good to merge.

No more, thank you. It looks pretty solid

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #514, opened 2026-05-08
fix(restore): record engine restore start failures in status

derekbit:

Issue Longhorn/longhorn#13061

Expose engine backup restore start failures through restore status so callers
can observe errors that happen before the restore goroutine is launched.

Preserve the last successfully restored backup across start failures while
still validating incremental restore eligibility before reusing it. Avoid
overwriting restore status for already-restored requests, and ignore generated
cscope files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ bin/
 .idea/
 *.iml
 *.ipr
+
+# cscope
+cscope.out*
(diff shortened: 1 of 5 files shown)


derekbit:
- -vs -m v2_volume_test test_basic.py::test_delete_backup_during_restoring_volume --count=5
  https://ci.longhorn.io/job/private/job/longhorn-tests-regression/10717/
- -vs -m v2_volume_test
  https://ci.longhorn.io/job/private/job/longhorn-tests-regression/10716/

davidcheng0922 (dismissed):
LGTM

davidcheng0922 (approved):
LGTM

derekbit:
@mergify backport v1.12.x

PR #518, opened 2026-05-11
fix: enforce snapshot-max-count/size for v2 volumes

carterli0407-cell:
The v2 (SPDK) engine had no enforcement logic for SnapshotMaxCount and SnapshotMaxSize. Add:
- SnapshotMaxCount/SnapshotMaxSize fields to Engine and Replica structs
- Enforcement check in SnapshotCreate
- VolumeSnapshotMaxCountSet/VolumeSnapshotMaxSizeSet gRPC handlers
- Client methods for the new RPCs

Fixes longhorn#12921


derekbit:
In favor of https://github.com/longhorn/longhorn-spdk-engine/pull/521

PR #521, opened 2026-05-14
feat: implement snapshot max count enforcement

davidcheng0922:

https://github.com/longhorn/longhorn/issues/12921


pkg/spdk/engine.go
```diff
@@ -1631,6 +1647,10 @@ func (e *Engine) snapshotOperationPreCheckWithoutLock(replicaClients map[string]
 		snapshotName = util.UUID()[:8]
 	}
 
+	if snapshotOp == SnapshotOperationCreate && e.SnapshotMaxCount > 0 && int32(len(e.SnapshotMap)) >= e.SnapshotMaxCount {
+		return "", fmt.Errorf("snapshot count %d is equal or larger than snapshotMaxCount %d", len(e.SnapshotMap), e.SnapshotMaxCount)
+	}
```
derekbit, line 1653: Should we skip the check for the system-generated snapshot?

davidcheng0922: To align with `v1` behavior, `v2` should also count and enforce snapshot max count for internal snapshots, such as replica-add/rebuilding snapshots.

ref: https://github.com/longhorn/longhorn-engine/blob/master/pkg/controller/control.go#L448-L472

derekbit (approved):
LGTM

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #531, opened 2026-05-15
fix: prevent resource leaks in backup, restore, and backing image cleanup paths

derekbit:

Issue https://github.com/longhorn/longhorn/issues/13114

Harden the backup and restore open paths so partial setup does not leave NVMe initiators or exposed bdevs behind.

### DIFF
--- a/pkg/spdk/backup.go
+++ b/pkg/spdk/backup.go
@@ -224,6 +224,9 @@ func (b *Backup) OpenSnapshot(snapshotName, volumeName string) (err error) {
 		return errors.Wrapf(err, "failed to create NVMe initiator for snapshot lvol bdev %v", lvolName)
 	}
 	if _, err := i.StartNvmeTCPInitiator(b.IP, strconv.Itoa(int(b.Port)), false, true); err != nil {
+		if _, stopErr := i.Stop(nil, true, true, true); stopErr != nil {
+			b.log.WithError(stopErr).Warnf("Failed to stop partially-started NVMe initiator for snapshot lvol bdev %v", lvolName)
+		}
 		return errors.Wrapf(err, "failed to start NVMe initiator for snapshot lvol bdev %v", lvolName)
 	}
 	b.initiator = i
--- a/pkg/spdk/backup.go
+++ b/pkg/spdk/backup.go
@@ -177,26 +177,29 @@ func (b *Backup) OpenSnapshot(snapshotName, volumeName string) (err error) {
 		if b.devFh != nil {
 			if errClose := b.devFh.Close(); errClose != nil {
 				b.log.WithError(errClose).Warnf("Failed to close NVMe device %v during backup open cleanup", b.devFh.Name())
+			} else {
+				b.devFh = nil
 			}
-			b.devFh = nil
 		}
 
 		if b.initiator != nil {
 			if _, stopErr := b.initiator.Stop(nil, true, true, true); stopErr != nil {
 				b.log.WithError(stopErr).Warnf("Failed to stop NVMe initiator for snapshot lvol bdev %v during backup open cleanup", lvolName)
+			} else {
+				b.initiator = nil
 			}
-			b.initiator = nil
 		}
 
 		if exposed {
 			if stopErr := backupStopExposeBdev(b.spdkClient, helpertypes.GetNQN(lvolName)); stopErr != nil {
 				b.log.WithError(stopErr).Warnf("Failed to unexpose snapshot lvol bdev %v during backup open cleanup", lvolName)
+			} else {
+				b.subsystemNQN = ""
+				b.controllerName = ""
 			}
 		}
 
 		b.fragmap = nil
-		b.subsystemNQN = ""
-		b.controllerName = ""
 	}()
 
 	b.replica.Lock()
@@ -311,6 +314,8 @@ func (b *Backup) CloseSnapshot(snapshotName, volumeName string) error {
 			} else {
 				errs = append(errs, errors.Wrap(err, "failed to close NVMe device"))
 			}
+		} else {
+			b.devFh = nil
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


davidcheng0922:
One question here, the restore (replica-level) is not used anymore, we use engine restore instead. Is this pr for backport to previous version?

derekbit:
> One question here, the restore (replica-level) is not used anymore, we use engine restore instead. Is this pr for backport to previous version?

No, we won't backport the PR.

davidcheng0922:
> > One question here, the restore (replica-level) is not used anymore, we use engine restore instead. Is this pr for backport to previous version?
> 
> No, we won't backport the PR.

Got it, just for reminder,  I think the changes in `replica.go` and `restore.go` are unreachable.

derekbit:
> > > One question here, the restore (replica-level) is not used anymore, we use engine restore instead. Is this pr for backport to previous version?
> > 
> > 
> > No, we won't backport the PR.
> 
> Got it, just for reminder, I think the changes in `replica.go` and `restore.go` are unreachable.

Could we put the fixes there and we can remove the unused APIs together in the future?

pkg/spdk/backup.go
```diff
@@ -222,33 +278,98 @@ func (b *Backup) CloseSnapshot(snapshotName, volumeName string) error {
...
+	}
+
+	if lvolName != "" {
+		b.log.Info("Unexposing snapshot lvol bdev")
+		if err := backupStopExposeBdev(b.spdkClient, helpertypes.GetNQN(lvolName)); err != nil {
+			errs = append(errs, errors.Wrapf(err, "failed to unexpose snapshot lvol bdev %v", lvolName))
+		}
 	}
 
+	b.releaseHeavyResourcesLocked()
```
davidcheng0922: Question: if `b.initiator.Stop()` or `StopExposeBdev()` fails, we still call `releaseHeavyResourcesLocked()`. 

That releases the port back to the allocator and clears the remaining cleanup state, but the NVMe initiator or exposed target may still exist. Could this cause a resource leak or even port reuse while the old target is still active?

pkg/spdk/backup.go
```diff
@@ -130,36 +155,67 @@ func (b *Backup) HasSnapshot(snapshotName, volumeName string) bool {
...
+		if exposed {
+			if stopErr := backupStopExposeBdev(b.spdkClient, helpertypes.GetNQN(lvolName)); stopErr != nil {
+				b.log.WithError(stopErr).Warnf("Failed to unexpose snapshot lvol bdev %v during backup open cleanup", lvolName)
+			}
+		}
+
+		b.fragmap = nil
+		b.subsystemNQN = ""
+		b.controllerName = ""
+	}()
```
davidcheng0922, line 203: Question: We assign `b.initiator` only after `StartNvmeTCPInitiator()` succeeds, and the deferred cleanup relies on `b.initiator` for rollback. 

If `StartNvmeTCPInitiator()` itself fails after partially creating the connection/path, the defer path may not stop that initiator instance. Do we need to handle this failure path here as well?

derekbit:
@davidcheng0922 Created 2 commits to fix the leak resources. Could you check them again? Thanks.

pkg/spdk/io_hooks.go
```diff
@@ -0,0 +1,85 @@
...
+	"os"
+
+	commonnet "github.com/longhorn/go-common-libs/net"
+	commonns "github.com/longhorn/go-common-libs/ns"
+	"github.com/longhorn/go-spdk-helper/pkg/initiator"
+	spdkclient "github.com/longhorn/go-spdk-helper/pkg/spdk/client"
+	spdktypes "github.com/longhorn/go-spdk-helper/pkg/spdk/types"
+	helperutil "github.com/longhorn/go-spdk-helper/pkg/util"
+
+	lhclient "github.com/longhorn/longhorn-spdk-engine/pkg/client"
```
davidcheng0922, line 14: NIT: seems need to re-organize

pkg/spdk/server_replica.go
```diff
@@ -15,6 +17,7 @@ import (
 	grpcstatus "google.golang.org/grpc/status"
 
 	"github.com/longhorn/backupstore"
+	btypes "github.com/longhorn/backupstore/types"
 	"github.com/longhorn/types/pkg/generated/spdkrpc"
 
 	butil "github.com/longhorn/backupstore/util"
```
davidcheng0922, line 23: NIT:  Re-organize the import

pkg/spdk/resource_cleanup_test.go
```diff
@@ -0,0 +1,409 @@
...
+	commonbitmap "github.com/longhorn/go-common-libs/bitmap"
+	commonns "github.com/longhorn/go-common-libs/ns"
+	"github.com/sirupsen/logrus"
+
+	"github.com/longhorn/go-spdk-helper/pkg/initiator"
+	spdkclient "github.com/longhorn/go-spdk-helper/pkg/spdk/client"
+	spdktypes "github.com/longhorn/go-spdk-helper/pkg/spdk/types"
+	helpertypes "github.com/longhorn/go-spdk-helper/pkg/types"
+
+	safelog "github.com/longhorn/longhorn-spdk-engine/pkg/log"
```
davidcheng0922: NIT: Re-organize the import

pkg/spdk/backup_test.go
```diff
@@ -0,0 +1,465 @@
...
+
+import (
+	"errors"
+	"time"
+
+	btypes "github.com/longhorn/backupstore/types"
+	commonbitmap "github.com/longhorn/go-common-libs/bitmap"
+	commonns "github.com/longhorn/go-common-libs/ns"
+	spdkclient "github.com/longhorn/go-spdk-helper/pkg/spdk/client"
+	"github.com/sirupsen/logrus"
```
davidcheng0922, line 11: NIT: Re-organize the import

derekbit:
@davidcheng0922 All are reorganized. Thank you.

davidcheng0922 (approved):
LGTM

derekbit:
@mergify backport v1.12.x

PR #533, opened 2026-05-18
fix(backup): harden backup failure handling

derekbit:

Issue Longhorn/longhorn#12842

Check replica state before reading snapshot data so backups fail through the normal error path when the replica is no longer running.

Also make UpdateBackupStatus preserve error as the final state even when progress reaches 100, and add tests to cover both error-first and error-after-completion cases.


derekbit:
@mergify backport v1.12.x

davidcheng0922 (dismissed):
LGTM

davidcheng0922 (approved):

davidcheng0922 (approved):

mantissahz (approved):
LGTM
And derekbit/longhorn-instance-manager:issue-12842-03 works as expected.

PR #539, opened 2026-05-24
fix(spdk): make engine frontend recovery async to prevent pod crash loop

derekbit:

Issue longhorn/longhorn#13185

When an instance-manager pod restarts with persisted engine frontend records referencing unreachable NVMe-TCP targets, the synchronous recoverEngineFrontends() blocks gRPC startup, exceeding the liveness probe threshold and causing an infinite pod create/delete loop.

The fix runs recovery as a goroutine so gRPC serves immediately. A TCP dial pre-check (5s timeout) in the recovery path detects unreachable targets early, avoiding the full 60s nvme-connect retry loop. Since recovery now runs concurrently with incoming gRPC requests, EngineFrontendCreate may race with the recovery goroutine for the same volume. To handle this safely, Create checks whether a conflicting map entry is still in Pending state (i.e. mid-recovery) and if so marks it Terminating and removes it from the map — letting the recovery goroutine perform the actual resource teardown sequentially after RecoverFromHost returns, which avoids concurrent access to the NVMe initiator. The recovery loop itself verifies map ownership (pointer equality) before deleting entries in both the failure and success paths, so it never accidentally removes a new frontend that was registered by a concurrent Create under the same name or volume.

### DIFF
--- a/pkg/spdk/enginefrontend.go
+++ b/pkg/spdk/enginefrontend.go
@@ -15,18 +15,18 @@ import (
 	"github.com/sirupsen/logrus"
 	"go.uber.org/multierr"
 
+	"github.com/longhorn/go-spdk-helper/pkg/initiator"
+	"github.com/longhorn/types/pkg/generated/spdkrpc"
+
+	commonbitmap "github.com/longhorn/go-common-libs/bitmap"
+	spdkclient "github.com/longhorn/go-spdk-helper/pkg/spdk/client"
+	helpertypes "github.com/longhorn/go-spdk-helper/pkg/types"
+
 	"github.com/longhorn/longhorn-spdk-engine/pkg/client"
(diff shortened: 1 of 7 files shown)


derekbit:
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7964/
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7965/

https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7966/
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7967

https://ci.longhorn.io/job/private/job/longhorn-tests-regression/10894/
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/10895/

derekbit:
cc @davidcheng0922 for review.

derekbit:
@mergify backport v1.12.x

davidcheng0922 (approved):
LGTM

PR #544, opened 2026-05-26
fix(recovery): do not tear down host initiator/dm resources on recovery failure

derekbit:

Issue Longhorn/longhorn#13185

Previously, recoverEngineFrontends called ef.Delete() when recovery failed or the frontend was superseded by a concurrent Create. Delete() invokes initiator.Stop() which disconnects NVMe controllers and removes device-mapper devices — potentially destroying resources that still serve active I/O for the new frontend (same subsystem NQN).

### DIFF
--- a/pkg/spdk/server.go
+++ b/pkg/spdk/server.go
@@ -3,6 +3,7 @@ package spdk
 import (
 	"context"
 	"fmt"
+	"os"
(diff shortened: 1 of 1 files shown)


derekbit:
@mergify backport v1.12.x

PR #547, opened 2026-05-28
fix: ensure isBDF() matches only whole strings

tserong:
Issue https://github.com/longhorn/longhorn/issues/13228

Add a V2/block disk using a `/dev/disk/by-path/pci-*` path, e.g. `/dev/disk/by-path/pci-0000:00:10.0-nvme-1`. Longhorn will incorrectly think that this is a BDF path, because it sees `0000:00:10.0` in the middle, and will fail to add the disk.

We can fix this by tightening up the regex in `isBDF()` to match only whole strings.

### DIFF
--- a/pkg/spdk/disk/types.go
+++ b/pkg/spdk/disk/types.go
@@ -124,7 +124,7 @@ func getDiskDriverForPath(diskDriver commontypes.DiskDriver, diskPath string) (c
 }
 
 func isBDF(addr string) bool {
-	bdfFormat := "[a-f0-9]{4}:[a-f0-9]{2}:[a-f0-9]{2}\\.[a-f0-9]{1}"
+	bdfFormat := "^[a-f0-9]{4}:[a-f0-9]{2}:[a-f0-9]{2}\\.[a-f0-9]{1}$"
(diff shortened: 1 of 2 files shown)


pkg/spdk/disk/types.go
```diff
@@ -124,7 +124,7 @@ func getDiskDriverForPath(diskDriver commontypes.DiskDriver, diskPath string) (c
 }
 
 func isBDF(addr string) bool {
-	bdfFormat := "[a-f0-9]{4}:[a-f0-9]{2}:[a-f0-9]{2}\\.[a-f0-9]{1}"
+	bdfFormat := "^[a-f0-9]{4}:[a-f0-9]{2}:[a-f0-9]{2}\\.[a-f0-9]{1}$"
```
derekbit, line 127: LGTM. Could you create a unit test case for `isBDF()`? Thank you.

tserong: Done!

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #552, opened 2026-05-29
refactor: linked clone feature

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/12552


derekbit:
Hello @shuo-wu 
Could we start reviewing the PR?

go.mod
```diff
@@ -2,18 +2,19 @@ module github.com/longhorn/longhorn-spdk-engine
 
 go 1.26.0
 
+replace github.com/longhorn/types => github.com/shuo-wu/types v0.0.0-20260602205913-cacbc769bf4c
+
```
derekbit: The PR is merged. We can remove the `replace`.

Dockerfile
```diff
@@ -49,8 +50,8 @@ RUN echo "Cloning longhorn/dep-versions SRC_BRANCH=${SRC_BRANCH} SRC_TAG=${SRC_T
     echo "dep-versions commit: $(git rev-parse HEAD)"
 
 # Build spdk
-RUN export REPO_OVERRIDE="" && \
-    export COMMIT_ID_OVERRIDE="" && \
+RUN export REPO_OVERRIDE="https://github.com/shuo-wu/spdk" && \
+    export COMMIT_ID_OVERRIDE="0873ff4bf" && \
```
derekbit: spdk PR is merged and versions.json in dep-versions is updated to the latest SPDK. We can remove the replacement. 
Ref: https://github.com/longhorn/dep-versions/blob/master/versions.json#L14-L18

pkg/spdk/replica.go
```diff
@@ -2230,23 +2604,57 @@ func (r *Replica) SnapshotCloneDstStatusCheck() (status *spdkrpc.ReplicaSnapshot
...
+			set, setParentErr = spdkClient.BdevLvolSetParent(r.Head.Alias, epAlias)
+			if setParentErr == nil {
+				break
+			}
+			if !strings.Contains(setParentErr.Error(), "busy") || i == maxSetParentRetries-1 {
+				return setParentErr
+			}
+			r.log.WithError(setParentErr).Debugf("BdevLvolSetParent returned busy (attempt %d/%d), retrying", i+1, maxSetParentRetries)
+			time.Sleep(time.Duration(i+1) * 200 * time.Millisecond)
+		}
```
davidcheng0922: NIT: Would it be better to keep the retry pattern consistent by using `retrygo.Do`(`github.com/avast/retry-go/v4`) instead of the hand-written retry loop?

pkg/spdk/replica.go
```diff
@@ -3924,6 +4645,23 @@ func (r *Replica) RebuildingDstSnapshotCreate(spdkClient *spdkclient.Client, sna
...
+			epAlias := spdktypes.GetLvolAlias(r.LvsName, epLvolName)
+			if _, err := spdkClient.BdevLvolSetParent(epAlias, snapSvcLvol.Alias); err != nil {
+				r.log.WithError(err).Warnf("Rebuilding dst replica failed to re-parent clone entrypoint %s to rebuilt snapshot %s; syncCloneReplicaInfo will repair on next startup", epLvolName, snapSvcLvol.Name)
+			} else {
+				r.log.Infof("Rebuilding dst replica re-parented clone entrypoint %s to rebuilt snapshot %s", epLvolName, snapSvcLvol.Name)
+			}
+		}
+		delete(r.rebuildingDstCache.detachedCloneEPsBySnap, snapshotName)
+	}
+
```
davidcheng0922, line 4764: Question about the failed reparent handling:

I saw the code comment saying that re-parent failures are non-fatal and syncCloneReplicaInfo() will repair the orphaned entrypoint at the next clone replica startup.

My concern is the window before that repair happens. If the entrypoint remains orphaned after `BdevLvolSetParent()` fails, is the clone lvol expected to keep serving I/O during this period? If so, do we still expect the data to be readable and correct?

If the data may not be reliable during this window, would it make sense to temporarily mark the affected clone replica as failed and transition it back to running once the recovery succeeds?

pkg/spdk/replica.go
```diff
@@ -2040,9 +2391,29 @@ func (r *Replica) SnapshotCloneDstStart(spdkClient *spdkclient.Client, snapshotN
...
+		r.isCloneReplica = true
+		r.cloneSourceReplicaName = srcReplicaName
+		r.cloneEntrypointLvolName = GetCloneEntrypointLvolName(srcReplicaName, snapshotName)
+
+		r.log.Infof("Dst replica starting linked-clone: src replica %s (addr %s), snapshot %s, entrypoint %s", srcReplicaName, srcReplicaAddress, snapshotName, r.cloneEntrypointLvolName)
 		r.snapshotCloningDstCache.cloningState = types.ProgressStateComplete
-		return r.SnapshotCloneDstFinish(spdkClient, cloneMode)
+		if err := r.SnapshotCloneDstFinish(spdkClient, cloneMode); err != nil {
+			return err
+		}
```
davidcheng0922, line 2495: Question about the linked-clone dst failure cleanup:

If `SnapshotCloneDstFinish()` fails after `ReplicaSnapshotCloneSrcStart()` succeeds, should we still best-effort call `ReplicaSnapshotCloneSrcFinish()` and clear `isSnapshotCloning` on the dst side?

Or is it intentional to keep the source-side clone cache in this case?

pkg/spdk/replica.go
```diff
@@ -3253,6 +3887,19 @@ func (r *Replica) RebuildingDstFinish(spdkClient *spdkclient.Client) (err error)
 			}
 		}
 
+		// Always perform cleanup to disconnect the external snapshot NVMe controller.
+		// Without this, subsequent ReplicaDelete would trigger bdev_nvme_detach_controller
+		// which can hang on same-node NVMe-oF connections.
+		_ = r.doCleanupForRebuildingDst(spdkClient)
```
davidcheng0922: NIT: Would it be helpful to add a debug log here instead of silently ignoring it?

pkg/spdk/engine.go
```diff
@@ -1672,6 +1717,22 @@ func (e *Engine) snapshotOperationPreCheckWithoutLock(replicaClients map[string]
...
+			if err != nil {
+				e.log.WithError(err).Warnf("Engine cannot verify clone-entrypoint guard for replica %s during snapshot %s deletion pre-check; skipping", replicaName, snapshotName)
+				continue
+			}
+			epLvolName := GetCloneEntrypointLvolName(replicaName, snapshotName)
+			if _, hasEP := r.CloneEntrypointMap[epLvolName]; hasEP {
+				return "", fmt.Errorf("cannot delete snapshot %s: replica %s has an active linked-clone entrypoint %s; delete the linked-clone volume first",
+					snapshotName, replicaName, epLvolName)
+			}
+		}
```
davidcheng0922, line 1733: ReplicaGet transient failed during pre-check
  -> engine skips clone-entrypoint guard for this replica
  -> engine sends SnapshotDelete to replica
  -> replica rejects because active clone entrypoint exists
  -> engine marks this replica ERR

Is it still the risk to  incorrectly marked ERR here?

derekbit (commented):
In general, LGTM except for the comments from @davidcheng0922. 
@shuo-wu Could you check the comments? Thakn you.

derekbit:
Hello @shuo-wu 
Could you resolve the conflicts and address the comments? We can resume the review of the PR.

go.mod
shuo-wu: The `type` update was already reverted. I will re-create a new PR later

pkg/spdk/engine.go
shuo-wu: Nice catch! When `replicaClient.ReplicaGet` returns an error, we should not continue here.

pkg/spdk/replica.go
shuo-wu: You are right, it's better to clean up the src side cache (even though `ReplicaSnapshotCloneSrcStart ()` fails) here. Will address this issue with a new commit. Plz take a look after the PR update!

pkg/spdk/replica.go
shuo-wu: > If the entrypoint remains orphaned after BdevLvolSetParent() fails, is the clone lvol expected to keep serving I/O during this period? If so, do we still expect the data to be readable and correct?

In my previous debugging and testing, I do find that orphan ep lvols keep serving I/O, which then leads to silent data corruption. But that's the buggy cases (I already fixed) and do not involve the rebuilding part.

> If the data may not be reliable during this window, would it make sense to temporarily mark the affected clone replica as failed and transition it back to running once the recovery succeeds?

Here, the current replica is being rebuilt, which means it was previously marked as ERR and all its clones are waiting for this rebuild to succeed. In other words, before this rebuilding is done, there will not be RW or even WO clone replicas there.

--- 

Maybe I can add a check (whether their parents are the corresponding snap lvols) for ep lvols in `validateAndUpdate`. This reparenting logic should be ok anyway.

derekbit:
Hello @shuo-wu 
CI failed

```
Running: golangci-lint
pkg/client/client_replica.go:518:5: QF1001: could apply De Morgan's law (staticcheck)
	if !(linkedCloneSrcReplicaName == "" && linkedCloneSrcEngineName == "" && linkedCloneSrcEngineAddress == "") &&
	   ^
pkg/client/client_replica.go:519:3: QF1001: could apply De Morgan's law (staticcheck)
		!(linkedCloneSrcReplicaName != "" && linkedCloneSrcEngineName != "" && linkedCloneSrcEngineAddress != "") {
		^
2 issues:
* staticcheck: 2
make: *** [Makefile:21: ci] Error 1
```

davidcheng0922 (commented):
Overall LGTM

derekbit (commented):
In general, LGTM.

shuo-wu:
Due to the refactoring introduced by the sharding PR merge, I need to recheck the entire PR and ensure my code adapts to the latest architecture.

shuo-wu:
This should be ready. It's already adapted to the latest implementation.

derekbit (dismissed):
In general, LGTM

davidcheng0922 (commented):
LGTM

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #568, opened 2026-06-11
fix(backup): retry opening NVMe device on ENXIO race

derekbit:

Issue Longhorn/longhorn#13188

After the NVMe-oF controller connects, the kernel may not have instantiated the namespace block device yet, causing open() on the device endpoint to return ENXIO. This intermittently fails backup creation right after StartNvmeTCPInitiator returns.

Retry the open up to 10 times with a 500ms fixed delay (5s total) when the error is ENXIO, using retry-go for consistency with the existing retry usage in pkg/spdk/engine.go. Non-ENXIO errors are wrapped in retrygo.Unrecoverable so they fail fast as before.

### DIFF
--- a/pkg/spdk/util.go
+++ b/pkg/spdk/util.go
@@ -3,9 +3,11 @@ package spdk
 import (
 	"fmt"
 	"net"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/avast/retry-go/v4"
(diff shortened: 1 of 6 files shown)


derekbit:
@mergify backport v1.12.x

c3y1huang (commented):
I am a bit confused - is this intended to fix  https://github.com/longhorn/longhorn/issues/13188?

I looked through the bundle in the issue description, but couldn't find `failed to open NVMe device` or `ENXIO` errors. Could you elaborate on how you arrived at this root cause? Is there evidence that links to this code path, or was it reproduced separately? 

Also, if this is the correct fix, should we apply the same handling to the other call site as well?

derekbit:
> I am a bit confused - is this intended to fix [longhorn/longhorn#13188](https://github.com/longhorn/longhorn/issues/13188)?
> 
> I looked through the bundle in the issue description, but couldn't find `failed to open NVMe device` or `ENXIO` errors. Could you elaborate on how you arrived at this root cause? Is there evidence that links to this code path, or was it reproduced separately?
> 
> Also, if this is the correct fix, should we apply the same handling to the other call site as well?

Yes. While working on the fix, I encountered an ENXIO error that caused backup failures.
As for the original issue in https://github.com/longhorn/longhorn/issues/13188, I am currently validating my another PR.
In summary, there are many issues that can result in backup failures.

c3y1huang:
> Yes. While working on the fix, I encountered an ENXIO error that caused backup failures.
> As for the original issue in [longhorn/longhorn#13188](https://github.com/longhorn/longhorn/issues/13188), I am currently validating my another PR.
> In summary, there are many issues that can result in backup failures.

Got it, so this PR is about ENXIO hardening. Should we do the same for the other call site as well?

pkg/spdk/util.go
```diff
@@ -23,6 +25,49 @@ import (
...
+				return openErr
+			}
+			return nil
+		},
+		retry.Attempts(openDeviceMaxRetries),
+		retry.Delay(openDeviceRetryInterval),
+		retry.DelayType(retry.FixedDelay),
+		retry.LastErrorOnly(true),
+		retry.OnRetry(func(n uint, err error) {
+			log.Warnf("NVMe device %v not ready (ENXIO), retrying (%d/%d)", endpoint, n+1, openDeviceMaxRetries)
```
c3y1huang, line 62: How about dropping the `(ENXIO)` and letting the errno come from the error itself (`.WithError(err)`)?

PR #592, opened 2026-07-09
fix(replica): handle stale SnapshotLvolMap after rebuild interrupt

shuo-wu:

Issue longhorn/longhorn#13315

When a rebuild is interrupted, the cached SnapshotLvolMap becomes stale. On the next Create() retry, construct() was skipped and getRootLvolName could non-deterministically pick an orphan lvol as root, causing validateAndUpdate() to fail with length mismatches.

Fix:
- Set reconstructRequired=true in RebuildingDstStart (before deleting snapshots) and in Delete() (when rebuild was interrupted). This ensures the next Create() re-scans SPDK state via construct().
- Broaden Create() reconstruct condition to also trigger for stopped replicas with reconstructRequired, not only pending ones.
- Rewrite getRootLvolName to walk backward from head via BaseSnapshot chain instead of scanning all lvols. This guarantees finding the root connected to the head, avoiding orphan selection.

### DIFF
--- a/pkg/spdk/replica.go
+++ b/pkg/spdk/replica.go
@@ -942,21 +942,32 @@ func (r *Replica) validateAndSyncLvstore(spdkClient *spdkclient.Client) error {
 	return nil
 }
 
-// getRootLvolName relies on the lvol name to identify if a lvol belongs to the replica,
-// then figuring out whether it is the root by checking the parent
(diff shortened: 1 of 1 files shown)


derekbit (approved):
LGTM

derekbit:
@mergify backport v1.12.x

PR #618, opened 2026-07-19
fix(backup): forward well-known backup parameters through v2 Labels

derekbit:

Issue Longhorn/longhorn#13526

The v2 (SPDK) BackupCreateRequest gRPC message has no dedicated field for backup parameters, so caller intent such as forcing a full backup ("backup-mode=full") is dropped between longhorn-instance-manager's proxy and longhorn-spdk-engine. As a result, v2 volumes stay Faulted after a "new full backup" attempt because backupstore's isFullBackup() sees an empty Parameters map and falls back to incremental.

Introduce a small pkg/types helper pair that smuggles well-known backup parameters through the existing Labels wire field under a reserved DNS-qualified prefix ("longhorn.io/backup-parameter."). The client wrappers EngineBackupCreate / ReplicaBackupCreate encode BackupCreateRequest.Parameters into Labels; ReplicaBackupCreate on the server side extracts them back into backupstore.DeltaBackupConfig. Parameters and removes them from the persisted user-visible labels.

The reserved namespace prevents a plain user label sharing a parameter's bare name (e.g. a caller tagging its backup with "backup-mode=full" for its own bookkeeping) from being mistaken for internal parameter intent. The engine-hop keeps forwarding labels opaquely, so the smuggled entries survive Engine.BackupCreate -> ReplicaBackupCreate without any special handling; a NOTE comment there documents this invariant.

### DIFF
--- a/pkg/client/client_engine.go
+++ b/pkg/client/client_engine.go
@@ -10,6 +10,7 @@ import (
 	"github.com/longhorn/types/pkg/generated/spdkrpc"
 
 	"github.com/longhorn/longhorn-spdk-engine/pkg/api"
+	"github.com/longhorn/longhorn-spdk-engine/pkg/types"
 	"github.com/longhorn/longhorn-spdk-engine/pkg/util"
 )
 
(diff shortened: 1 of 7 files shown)


derekbit:
e2e: https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8510

mantissahz (approved):
In general, LGTM

derekbit:
@mergify backport v1.12.x

PR #639, opened 2026-08-07
fix(clone): erase ep info for root snap of clone volumes

shuo-wu:
Longhorn 13671

Issue https://github.com/longhorn/longhorn/issues/13671


shuo-wu:
@mergify backport v1.12.x

shuo-wu:
Close this fix since it will impact the rebuild. This EP info returned by this API is required for rebuilding.

PR #642, opened 2026-08-09
refactor: refactor linked clone gRPC APIs and adjust the rebuild flow for linked clone

shuo-wu:

https://github.com/longhorn/longhorn/issues/13671

1. The linked clone fields introduced by the refactoring are messy. This PR is to make these APIs cleaner.
2. Avoid exposing the entrypoint info in the API response. The ep info is internal only for refactored linked clone volumes.
3. Refactor the rebuilding dst so that the rebuilding dst replica will directly get the clone src snapshot name from the caller (engine) rather than detecting it manually in the end


derekbit:
@shuo-wu Could you create a BUG ticket for the PR?

derekbit (commented):
In general, LGTM.
v1.12.1-rc4 will be released after validating this change.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/longhorn/longhorn-spdk-engine
 
 go 1.26.0
 
+replace github.com/longhorn/types => github.com/shuo-wu/types v0.0.0-20260809045332-36ecea94a60f
```
derekbit: We can update the go.mod since the longhorn/types PR is merged. Thank you.

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #644, opened 2026-08-11
feat(replica): drive the host-ACL restriction from the ReplicaCreate request

hookak:

Issue longhorn/longhorn#13699

Replica exposures follow the host-ACL flag carried by `ReplicaCreateRequest`, applied on every create including replicas reconstructed by the startup scan. Shard and shard group exposures keep the unconditional restriction.

Depends on the longhorn/types PR (go.mod bump after it merges). Merged alone this reverts replica exposures to any-host — ships together with the instance-manager and manager counterparts.

`None`

### DIFF
--- a/pkg/client/client_replica.go
+++ b/pkg/client/client_replica.go
@@ -15,7 +15,7 @@ import (
 )
 
 // ReplicaCreate creates and starts a replica in the specified lvstore.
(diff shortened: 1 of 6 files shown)


derekbit:
Hello @hookak Is the PR ready for review?

cc @davidcheng0922

PR #654, opened 2026-08-20
feat(spdk): thread NVMe-oF transport type through engine/replica create

abonillabeeche:
Part of longhorn/longhorn#13796. **Depends on longhorn/types#121 and longhorn/go-spdk-helper#308.**

Carry the `DataEngineTransport` from the spdkrpc create requests into replica and engine setup: create the matching NVMe-oF transport (RDMA or TCP) on the SPDK target, add the listener with the right trtype, and connect the engine's controllers over the selected transport. RDMA uses RoCEv2 (`adrfam=ipv4`). Unset means TCP; existing volumes and the host frontend are unaffected (frontend stays TCP).

Files: `pkg/spdk/{engine,replica,server,server_engine,util,shard,shardgroup}.go`, `pkg/client/{client_engine,client_replica}.go`.

### DIFF
--- a/pkg/client/client_engine.go
+++ b/pkg/client/client_engine.go
@@ -20,7 +20,7 @@ import (
 // DATA_LAYOUT_TYPE_SHARDED constructs a single shardGroupBackend (EC).
 // Without forwarding this field, callers would default to the proto3 zero
 // value (REPLICATED), and EC volumes would be silently miscreated as RAID1.
-func (c *SPDKClient) EngineCreate(name, volumeName, frontend string, specSize uint64, replicaAddressMap map[string]string, portCount int32, salvageRequested bool, snapshotMaxCount int32, dataLayoutType spdkrpc.DataLayoutType, transportType spdkrpc.DataEngineTransport) (*api.Engine, error) {
+func (c *SPDKClient) EngineCreate(name, volumeName, frontend string, specSize uint64, replicaAddressMap map[string]string, portCount int32, salvageRequested bool, snapshotMaxCount int32, dataLayoutType spdkrpc.DataLayoutType, transportType spdkrpc.TransportType) (*api.Engine, error) {
 	if name == "" {
 		return nil, fmt.Errorf("failed to start engine: missing required parameter name")
 	}
@@ -45,7 +45,7 @@ func (c *SPDKClient) EngineCreate(name, volumeName, frontend string, specSize ui
 		SalvageRequested:    salvageRequested,
 		SnapshotMaxCount:    snapshotMaxCount,
 		DataLayoutType:      dataLayoutType,
-		DataEngineTransport: transportType,
+		TransportType: transportType,
 	})
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to start engine")
--- a/pkg/client/client_engine.go
+++ b/pkg/client/client_engine.go
@@ -20,7 +20,7 @@ import (
 // DATA_LAYOUT_TYPE_SHARDED constructs a single shardGroupBackend (EC).
 // Without forwarding this field, callers would default to the proto3 zero
 // value (REPLICATED), and EC volumes would be silently miscreated as RAID1.
-func (c *SPDKClient) EngineCreate(name, volumeName, frontend string, specSize uint64, replicaAddressMap map[string]string, portCount int32, salvageRequested bool, snapshotMaxCount int32, dataLayoutType spdkrpc.DataLayoutType) (*api.Engine, error) {
+func (c *SPDKClient) EngineCreate(name, volumeName, frontend string, specSize uint64, replicaAddressMap map[string]string, portCount int32, salvageRequested bool, snapshotMaxCount int32, dataLayoutType spdkrpc.DataLayoutType, transportType spdkrpc.DataEngineTransport) (*api.Engine, error) {
 	if name == "" {
 		return nil, fmt.Errorf("failed to start engine: missing required parameter name")
 	}
@@ -36,15 +36,16 @@ func (c *SPDKClient) EngineCreate(name, volumeName, frontend string, specSize ui
 	defer cancel()
 
 	resp, err := client.EngineCreate(ctx, &spdkrpc.EngineCreateRequest{
-		Name:              name,
-		VolumeName:        volumeName,
-		Frontend:          frontend,
-		SpecSize:          specSize,
-		ReplicaAddressMap: replicaAddressMap,
-		PortCount:         portCount,
-		SalvageRequested:  salvageRequested,
-		SnapshotMaxCount:  snapshotMaxCount,
-		DataLayoutType:    dataLayoutType,
+		Name:                name,
+		VolumeName:          volumeName,
+		Frontend:            frontend,
+		SpecSize:            specSize,
+		ReplicaAddressMap:   replicaAddressMap,
+		PortCount:           portCount,
+		SalvageRequested:    salvageRequested,
+		SnapshotMaxCount:    snapshotMaxCount,
+		DataLayoutType:      dataLayoutType,
+		DataEngineTransport: transportType,
 	})
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to start engine")
--- a/pkg/spdk/engine.go
+++ b/pkg/spdk/engine.go
@@ -124,6 +124,14 @@ type Engine struct {
 	ActualSize uint64
 	Frontend   string
 
+	// TransportType is the NVMe-oF transport this engine uses to connect to its
+	// replicas' exposed head bdevs. It must match the transport each replica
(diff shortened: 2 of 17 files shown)


pkg/client/client_engine.go
```diff
@@ -36,15 +36,16 @@ func (c *SPDKClient) EngineCreate(name, volumeName, frontend string, specSize ui
...
+		Name:                name,
+		VolumeName:          volumeName,
+		Frontend:            frontend,
+		SpecSize:            specSize,
+		ReplicaAddressMap:   replicaAddressMap,
+		PortCount:           portCount,
+		SalvageRequested:    salvageRequested,
+		SnapshotMaxCount:    snapshotMaxCount,
+		DataLayoutType:      dataLayoutType,
+		DataEngineTransport: dataEngineTransport,
```
shuo-wu: NIT: `TransportType` may be better. No need to add prefix `DataEngine`

pkg/spdk/engine.go
```diff
@@ -124,6 +124,14 @@ type Engine struct {
 	ActualSize uint64
 	Frontend   string
 
+	// Trtype is the NVMe-oF transport this engine uses to connect to its
+	// replicas' exposed head bdevs. It must match the transport each replica
+	// used to expose (Replica.Trtype). Set from EngineCreateRequest.
+	// data_engine_transport; unset (empty) is treated as TCP for backward
+	// compatibility. Only the internal engine<->replica data fabric honors this;
+	// the host-facing NVMe-TCP frontend and all EC/transient paths stay TCP.
+	Trtype spdktypes.NvmeTransportType
```
shuo-wu: NIT: `TransportType`

pkg/spdk/engine.go
```diff
@@ -365,7 +376,7 @@ func (e *Engine) createNVMeTCPTarget(spdkClient *spdkclient.Client, superiorPort
 	e.log.Infof("Starting to expose RAID bdev for engine target %v on %v:%v with initial ANA state %v, cntlid %v, nsUUID %v",
 		e.Name, e.NvmeTcpTarget.IP, e.NvmeTcpTarget.Port, initialANAState, cntlid, nsUUID)
 	if err := spdkClient.StartExposeBdevWithANAState(e.NvmeTcpTarget.Nqn, e.Name, e.NvmeTcpTarget.Nguid, nsUUID,
-		e.NvmeTcpTarget.IP, strconv.Itoa(int(e.NvmeTcpTarget.Port)), spdkANAState, cntlid, cntlid); err != nil {
+		e.NvmeTcpTarget.IP, strconv.Itoa(int(e.NvmeTcpTarget.Port)), spdktypes.NvmeTransportTypeTCP, spdkANAState, cntlid, cntlid); err != nil {
```
shuo-wu, line 382: NIT: Add a `TODO` comment for the Frontend adaptation

pkg/spdk/engine.go
```diff
@@ -387,7 +398,7 @@ func (e *Engine) connectReplicas(spdkClient *spdkclient.Client, replicaAddressMa
 	for replicaName, replicaAddr := range replicaAddressMap {
 		e.backends[replicaName] = backendFactory(replicaName, replicaAddr)
 
-		bdevName, err := connectNVMfBdev(spdkClient, replicaName, replicaAddr, e.ctrlrLossTimeout, e.fastIOFailTimeoutSec, maxRetries, retryInterval)
+		bdevName, err := connectNVMfBdev(spdkClient, replicaName, replicaAddr, e.Trtype, e.ctrlrLossTimeout, e.fastIOFailTimeoutSec, maxRetries, retryInterval)
```
shuo-wu: Probably we need to validate whether the replicas are using the same tytype with this engine. The validation can be applied by refactoring func `validateReplicaSize`

pkg/spdk/server_engine.go
```diff
@@ -43,6 +43,11 @@ func (s *Server) EngineCreate(ctx context.Context, req *spdkrpc.EngineCreateRequ
 		e = s.engineMap[req.Name]
 	}
 
+	// Select the internal engine<->replica NVMe-oF transport. Unset (TCP=0)
+	// preserves historical TCP behavior. Must match the transport the replicas
+	// used to expose their heads.
+	e.Trtype = nvmeTransportFromProto(req.DataEngineTransport)
+
```
shuo-wu: It's better to directly add an input `TransportType` for the above `NewEngine` instead

pkg/spdk/replica.go
```diff
@@ -2532,7 +2539,7 @@ func (r *Replica) SnapshotCloneDstStart(spdkClient *spdkclient.Client, snapshotN
 
 	if err := spdkClient.StartExposeBdev(helpertypes.GetNQN(r.snapshotCloningDstCache.cloningLvol.Name),
 		r.snapshotCloningDstCache.cloningLvol.UUID, generateNGUID(r.snapshotCloningDstCache.cloningLvol.Name), r.IP,
-		strconv.Itoa(int(r.snapshotCloningDstCache.cloningPort)), helpertypes.InternalHostNQN); err != nil {
+		strconv.Itoa(int(r.snapshotCloningDstCache.cloningPort)), spdktypes.NvmeTransportTypeTCP, helpertypes.InternalHostNQN); err != nil {
```
shuo-wu, line 2545: Can we use RDMA here (clone) or in the below rebuilding/expansion? Or you can add `TODO` comments first.

shuo-wu:
Besides, I am thinking whether Longhorn needs to detect the availability of RDMA hardware before accepting the transport type `RDMA`. 
The detection can be in this spdk-server or in the upper longhorn-manager. Currently, I may prefer to let `longhorn-manager` detect that. Then, Longhorn can reject the invalid engine earlier. Besides, if only some of the nodes have RDMA hardware, the validation will be more complicated.

abonillabeeche:
Thanks for the review — pushed a revision addressing the NITs.

- **`engine.go:133` / field naming** — renamed the struct field to `TransportType` on both `Engine` and `Replica`, and the client-method params to `transportType`.
- **`server_engine.go:50`** — `TransportType` is now an argument to `NewEngine(...)` and set at construction; dropped the post-construct assignment.
- **`engine.go:379` (frontend) and `replica.go:2542` (clone/rebuild/expansion)** — added `TODO`s. These transient/host-facing paths stay on NVMe-TCP for this MVP; the internal engine↔replica head fabric is the only transport-selectable path. Honoring RDMA on the frontend also needs the initiator/host side to support it, so I'd rather do it as a follow-up than half-wire it here.
- **`client_engine.go:48` — dropping the `DataEngine` prefix**: the remaining `DataEngineTransport` here is the generated proto field from longhorn/types#121. Renaming it to `transport_type` is a clean change but cascades to longhorn-manager and instance-manager. Want me to rename it in #121 for consistency, or keep the proto field namespaced and only use `TransportType` on the Go structs?
- **`engine.go:401` — validating engine and replicas agree on trtype**: good idea. Right now `validateNvmeTransport` accepts TCP or RDMA but doesn't cross-check that a replica's transport matches the engine's, because the engine doesn't yet see each replica's declared transport at connect time. I'll do this as a follow-up: surface the replica transport (via `BackendView`) and add the equality check by refactoring `validateReplicaSize` as you suggested. Flagging it on the tracking issue.
- **RDMA hardware detection**: agreed it belongs in longhorn-manager rather than here — it can reject an RDMA engine before scheduling, and it's the right place to reason about mixed clusters where only some nodes have RoCE-capable NICs. I'll add node-level RDMA-capability detection + validation on the manager side (longhorn/longhorn-manager#5107, tracked in longhorn/longhorn#13796) rather than in the spdk-server.

Default behavior is unchanged: an unset transport maps to TCP, so an older control plane keeps working. Not e2e-tested in this revision.

pkg/spdk/server_engine.go
abonillabeeche: TransportType is now an argument to NewEngine(...) and set at construction; dropped the post-construct assignment.

pkg/spdk/engine.go
abonillabeeche: renamed the struct field to TransportType on both Engine and Replica, and the client-method params to transportType

pkg/client/client_engine.go
abonillabeeche: dropping the DataEngine prefix: the remaining DataEngineTransport here is the generated proto field from https://github.com/longhorn/types/pull/121. Renaming it to transport_type is a clean change but cascades to longhorn-manager and instance-manager. Want me to rename it in https://github.com/longhorn/longhorn-spdk-engine/pull/121 for consistency, or keep the proto field namespaced and only use TransportType on the Go structs?

pkg/spdk/replica.go
abonillabeeche: added TODOs.

PR #664, opened 2026-08-29
fix: recover a v2 block disk whose device is not driven by the kernel

derekbit:

Issue Longhorn/longhorn#13893

An interrupted disk creation leaves the PCI device without a usable kernel driver, either bound to a userspace PCI driver or bound to nothing at all when the userspace bind failed after the kernel driver was released. Neither state exposes a block device, so driver detection for "auto" failed and the disk stayed Ready=False and Schedulable=False forever.
    
Resolve the driver from the PCI device type reported by the SPDK setup script in that state, and release the device instead of leaking it: on a failed driver detection, and on a deletion where the instance manager has no record of the disk, which happens after a restart because the record only lives in memory. A disk in the error state is now retried instead of latching forever, and its failure reason is reported through the new Disk.message field.


pkg/spdk/disk/cleanup.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	if !IsDetachedFromKernelDriver(diskStatus.Driver) {
+		return false, nil
+	}
+
+	diskDriver := commontypes.DiskDriverNvme
+	if strings.EqualFold(diskStatus.Type, PciDeviceTypeVirtio) {
+		diskDriver = commontypes.DiskDriverVirtioBlk
+	}
```
davidcheng0922: NIT: Do we have to handle unknown type? `getDriverForDetachedDevice` returns an error for the same input, so the two functions disagree on the same diskStatus.

derekbit: Good catch, fixed. The unknown type now returns an error, so both functions agree on it.

pkg/spdk/disk/nvme/nvme.go
```diff
@@ -60,12 +62,19 @@ func (d *DiskDriverNvme) DiskCreate(spdkClient *spdkclient.Client, diskName, dis
 		return "", errors.Wrapf(err, "failed to attach NVMe disk %v", diskPath)
 	}
 	if len(bdevs) == 0 {
-		return "", errors.Errorf("failed to attach NVMe disk %v", diskPath)
+		err = errors.Errorf("no bdev is created after attaching NVMe disk %v", diskPath)
+		return "", err
```
davidcheng0922: NIT: Consider returning the error directly to keep the style consistent.

pkg/spdk/disk/virtio-blk/virtio-blk.go
```diff
@@ -50,7 +52,8 @@ func (d *DiskDriverVirtioBlk) DiskCreate(spdkClient *spdkclient.Client, diskName
 		return "", errors.Wrapf(err, "failed to attach virtio-blk disk %v", diskPath)
 	}
 	if len(bdevs) == 0 {
-		return "", errors.Errorf("failed to attach virtio-blk disk %v", diskPath)
+		err = errors.Errorf("no bdev is created after attaching virtio-blk disk %v", diskPath)
+		return "", err
```
davidcheng0922: NIT: Same

pkg/spdk/disk/virtio-scsi/virtio-scsi.go
```diff
@@ -50,7 +52,8 @@ func (d *DiskDriverVirtioScsi) DiskCreate(spdkClient *spdkclient.Client, diskNam
 		return "", errors.Wrapf(err, "failed to attach virtio-scsi disk %v", diskPath)
 	}
 	if len(bdevs) == 0 {
-		return "", errors.Errorf("failed to attach virtio-scsi disk %v", diskPath)
+		err = errors.Errorf("no bdev is created after attaching virtio-scsi disk %v", diskPath)
+		return "", err
```
davidcheng0922: NIT: Same

davidcheng0922 (commented):
Overall, LGTM.

pkg/spdk/disk/nvme/nvme.go
derekbit: Updated. Thank you.

pkg/spdk/disk/virtio-scsi/virtio-scsi.go
derekbit: Updated. Thank you.

pkg/spdk/disk/virtio-blk/virtio-blk.go
derekbit: Updated. Thank you.

derekbit:
@mergify backport v1.13.x

davidcheng0922 (approved):
LGTM

derekbit:
@mergify backport v1.13.x

derekbit:
@mergify backport v1.12.x

PR #674, opened 2026-09-02
fix: keep the failure reason of a v2 block disk across a retried creation

derekbit:

Issue Longhorn/longhorn#13893

Retrying a failed creation replaces the Disk record so that no resolved value of the previous attempt, such as the disk ID, the driver or the sizes, survives into the new one. That also dropped ErrorMsg, which is not per-attempt state but the diagnostic history the caller relies on.

The retry then starts over in the creating state and only afterwards queues for the disk creation lock, which serializes the creation of every disk on the node. While it waits, DiskGet reports the creating state with an empty message, so a caller polling periodically loses the reason of the failure and reports a generic message instead. Once several disks are created at the same time, for instance after a node reboot, the wait spans multiple polls.

Carry ErrorMsg over to the new record so the previous reason is still reported until the retry reaches a conclusion. It is overwritten on a failure and cleared on a success, so a healthy disk never keeps a stale reason.

### DIFF
--- a/pkg/spdk/disk.go
+++ b/pkg/spdk/disk.go
@@ -76,6 +76,12 @@ func (d *Disk) GetState() DiskState {
 	return d.State
 }
 
+func (d *Disk) GetErrorMsg() string {
+	d.RLock()
+	defer d.RUnlock()
+	return d.ErrorMsg
+}
+
 func NewDisk(diskName, diskUUID, diskPath, diskDriver string, blockSize int64) *Disk {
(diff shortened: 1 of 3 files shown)


derekbit:
@mergify backport v1.13.x

derekbit:
@mergify backport v1.13.x

derekbit:
@mergify backport v1.12.x

davidcheng0922 (approved):
LGTM

PR #681, opened 2026-09-06
feat(network): IM-scope configurable ip family

COLDTURNIP:

Issue longhorn/longhorn#13050


pkg/spdk/server.go
```diff
@@ -1118,6 +1169,22 @@ func (s *Server) recoverEngineFrontends(ctx context.Context) {
...
 	}
 
+	validRecords := make([]*EngineFrontendRecord, 0, len(records))
+	for _, record := range records {
+		if err := s.validateEngineFrontendRecord(record); err != nil {
+			logrus.WithError(err).Warnf("Removing engine frontend %s from recovery", record.Name)
+			if removeErr := removeEngineFrontendRecord(s.metadataDir, record.VolumeName); removeErr != nil {
+				logrus.WithError(removeErr).Warnf("Failed to remove invalid engine frontend %s record during recovery", record.Name)
+			}
+			continue
```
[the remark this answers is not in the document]
COLDTURNIP: Backing image support for v2 engine is now deprecated and no longer maintained. Will remove this part.

pkg/spdk_test.go
```diff
@@ -195,7 +195,7 @@ func LaunchTestSPDKTargetDaemon(c *C, execute func(envs []string, name string, a
 func launchTestSPDKGRPCServer(ctx context.Context, c *C, ip string, execute func(envs []string, name string, args []string, timeout time.Duration) (string, error), wg *sync.WaitGroup) *server.Server {
 
 	LaunchTestSPDKTargetDaemon(c, execute)
-	srv, err := server.NewServer(ctx, defaultTestStartPort, defaultTestEndPort, nil)
+	srv, err := server.NewServer(ctx, defaultTestStartPort, defaultTestEndPort, commonnet.IPFamilyUnspecified, nil)
```
derekbit, line 198: Isn't the IPFamily determined by the setting from longhorn-manager? Why do we always set it to `IPFamilyUnspecified`?

COLDTURNIP: `launchTestSPDKGRPCServer` creates a loop-back TCP listener to serve gRPC API for testing (the IP is from `commonnet.GetAnyExternalIP()`), and the test environment may not support all kinds of IP family.

pkg/spdk/server.go
```diff
@@ -1118,6 +1169,22 @@ func (s *Server) recoverEngineFrontends(ctx context.Context) {
...
+		}
+		validRecords = append(validRecords, record)
+	}
+	records = validRecords
+	if len(records) == 0 {
+		return
+	}
```
derekbit: Why do we need to the validation? If user wants to change the IP family, they need to detach all volumes first. Ideally, there are no residual engine frontend and we don't need to check the IP families, isn't it?

pkg/spdk/enginefrontend.go
```diff
@@ -15,12 +15,12 @@ import (
...
 
-	"github.com/longhorn/go-spdk-helper/pkg/initiator"
-	"github.com/longhorn/types/pkg/generated/spdkrpc"
-
 	commonbitmap "github.com/longhorn/go-common-libs/bitmap"
+	commonnet "github.com/longhorn/go-common-libs/net"
+	"github.com/longhorn/go-spdk-helper/pkg/initiator"
 	spdkclient "github.com/longhorn/go-spdk-helper/pkg/spdk/client"
 	helpertypes "github.com/longhorn/go-spdk-helper/pkg/types"
+	"github.com/longhorn/types/pkg/generated/spdkrpc"
```
derekbit: Reorganize the imports

pkg/spdk/server_engine.go
```diff
@@ -11,6 +11,7 @@ import (
 	grpccodes "google.golang.org/grpc/codes"
 	grpcstatus "google.golang.org/grpc/status"
 
+	commonnet "github.com/longhorn/go-common-libs/net"
 	"github.com/longhorn/types/pkg/generated/spdkrpc"
```
derekbit: Reorganize the imports

pkg/spdk/server_engine.go
```diff
@@ -215,6 +218,11 @@ func (s *Server) EngineFrontendSwitchOver(ctx context.Context, req *spdkrpc.Engi
 	if targetIP, targetPort, splitErr := splitHostPort(req.TargetAddress); splitErr != nil || targetIP == "" || targetPort == 0 {
 		return nil, grpcstatus.Errorf(grpccodes.InvalidArgument, "invalid target address %q", req.TargetAddress)
 	}
+	if s.ipFamily != commonnet.IPFamilyUnspecified {
+		if familyErr := validatePersistedAddressFamily(req.TargetAddress, s.ipFamily); familyErr != nil {
+			return nil, grpcstatus.Errorf(grpccodes.InvalidArgument, "invalid target address %q: %v", req.TargetAddress, familyErr)
+		}
+	}
```
derekbit: Do we still need the validation if the IP family is global same?

pkg/spdk/server.go
```diff
@@ -1145,8 +1181,9 @@ func (s *Server) recoverEngineFrontends(ctx context.Context) {
 			continue
 		}
 
-		ef := NewEngineFrontend(record.Name, record.EngineName, record.VolumeName,
-			record.Frontend, record.SpecSize, 0, 0, s.updateChs[types.InstanceTypeEngineFrontend], s.newServiceClient)
+		ef := newEngineFrontend(record.Name, record.EngineName, record.VolumeName,
+			record.Frontend, record.SpecSize, 0, 0, s.ipFamily,
+			s.updateChs[types.InstanceTypeEngineFrontend], s.newServiceClient)
```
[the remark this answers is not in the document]
COLDTURNIP: IP-family changes require all volumes to be detached. Successful frontend deletion also removes the persisted record, so there should be no frontend to recover across a family change. Recovery of attached volumes after an IM restart uses the unchanged family. Records left after detachment are orphaned lifecycle metadata; their detection and cleanup should be addressed separately rather than expanding this feature's scope.

pkg/spdk/server.go
```diff
@@ -94,7 +100,35 @@ type Server struct {
 	newServiceClient ServiceClientFactory
 }
 
-func NewServer(ctx context.Context, portStart, portEnd int32, newServiceClient ServiceClientFactory) (*Server, error) {
+func validateIPFamily(family commonnet.IPFamily) error {
```
c3y1huang, line 103: Duplicated to commonnet.ParseIPFamily.

pkg/spdk/io_hooks.go
```diff
@@ -56,7 +56,7 @@ var (
 	restoreExposeSnapshotLvolBdev = exposeSnapshotLvolBdev
 	restoreStopExposeBdev         = func(cli *spdkclient.Client, nqn string) error { return cli.StopExposeBdev(nqn) }
 
-	backingImageGetIPForPod = commonnet.GetIPForPod
+	backingImageGetIPForPod = func() (string, error) { return commonnet.GetIPForPodByNetworkAndFamily(commonnet.IPFamilyUnspecified) }
```
c3y1huang, line 59: Slightly confused, does this mean backing image exposure doesn't follow the IM configured family? Same question for the backup.

COLDTURNIP: Backing image for v2 data engine is now deprecated, hence the minimal change is made here. To the backup, the IP information is copied from the replica.

PR #689, opened 2026-09-09
fix: resolve NVMe path teardown deadlock

derekbit:

Issue Longhorn/longhorn#13913


pkg/spdk/engine.go
```diff
@@ -359,24 +360,69 @@ func (e *Engine) createNVMeTCPTarget(spdkClient *spdkclient.Client, superiorPort
...
+	spdkANAState, err := toSPDKListenerANAState(currentANAState)
+	if err != nil {
+		return errors.Wrapf(err, "invalid ANA state %q for engine target %v during CNTLID rotation", currentANAState, e.Name)
+	}
+
+	e.log.Infof("Re-exposing engine target %v with next CNTLID %v after duplicate CNTLID", e.Name, nextCntlid)
+	if err := spdkClient.StopExposeBdev(e.NvmeTcpTarget.Nqn); err != nil && !jsonrpc.IsJSONRPCRespErrorNoSuchDevice(err) {
+		return errors.Wrapf(err, "failed to stop exposing bdev for engine target %v before CNTLID rotation", e.Name)
+	}
+	if err := e.startExposeNVMeTCPTargetLocked(spdkClient, currentANAState, spdkANAState, nextCntlid); err != nil {
```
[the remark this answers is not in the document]
derekbit: StopExposeBdev is subsystem-wide for the local SPDK instance, but I don't think this creates the multi-host interruption described here in the Longhorn-managed v2 frontend model.

RWX clients do not each connect to the NVMe/TCP subsystem. The share-manager owns the single block frontend and re-exports it over NFS, so RWX does not imply multiple NVMe hosts on the same target subsystem.

For migration/switchover, multiple EngineFrontend CRs can exist for a volume, but the targets live in separate SPDK instances. StopExposeBdev only removes the subsystem in the local SPDK instance, and the spdk-engine server also rejects a second non-pending EngineFrontend for the same volume in the same instance.

This path is only reached after a new frontend connect fails with Duplicate cntlid, before that frontend has a usable I/O path. It is meant to bypass a stale host-side controller, not rotate while other Longhorn-managed healthy writers are using the same local target subsystem.

pkg/spdk/enginefrontend.go
```diff
@@ -141,8 +147,45 @@ const (
...
+}
+
+func isTransientNvmeDeviceStateError(err error) bool {
+	for _, state := range transientNvmeDeviceStates {
+		if strings.Contains(err.Error(), state) {
+			return true
+		}
+	}
+	return false
+}
```
shuo-wu, line 193: NIT: Is it better to write the code as below?
```
const NVMeDeviceStateConnecting = "connecting"
......

const transientNvmeDeviceStates = []string {
	NVMeDeviceStateConnecting,
	......
}

func isTransientNvmeDeviceStateError(err error) bool {
	for _, state := range transientNvmeDeviceStates {
			if strings.Contains(err.Error(), fmt.Sprintf("%s state", state)) {
				return true
			}
	}
	return false
```

pkg/spdk/enginefrontend.go
```diff
@@ -2599,6 +2648,9 @@ func (ef *EngineFrontend) validateAndUpdateNvmeTcpFrontend() (err error) {
 			}
 			return err
 		}
+		if !ef.transientDeviceStateSince.IsZero() && time.Since(ef.lastTransientDeviceStateAt) >= transientDeviceStateClearInterval {
+			ef.forgetTransientDeviceState()
+		}
```
shuo-wu, line 2703: When the code is here, it means EngineFrontend successfully loaded a Nvme device with a **valid state**. Should we still need to wait an extra 30s to clean up the timestamp?

pkg/spdk/server_enginefrontend.go
```diff
@@ -307,6 +308,17 @@ func (s *Server) EngineFrontendCreate(ctx context.Context, req *spdkrpc.EngineFr
...
 
+	rotateTargetCntlid := func() error {
+		s.RLock()
+		engine := s.engineMap[req.EngineName]
+		spdkClient := s.spdkClient
+		s.RUnlock()
+		if engine == nil {
+			return fmt.Errorf("cannot find engine %v for target CNTLID rotation", req.EngineName)
+		}
+		return engine.ReExposeNVMeTCPTargetWithNextCntlid(spdkClient)
```
shuo-wu: Let the server handle this NVMe controller ID stuff looks weird. The server should not care about this detail.

How about using an interface inside `EngineFrontend` like `Engine.replicaAdder` instead? For regular NVMe EF, it will be set to `ReExposeNVMeTCPTargetWithNextCntlid` during `NewEngineFrontend`. And the test will change it to arbitrary implementations.

shuo-wu (commented):
In general LGTM

pkg/spdk/enginefrontend.go
derekbit: Good question. I kept the 30s clean-read interval on purpose, because a successful `loadInitiatorNVMeDeviceInfo` here doesn't always mean the frontend has actually recovered. In the stale-target case this PR targets (the IM pod is recreated with a new IP, so the target moves), validation flaps: one round can still read the namespace device while the next round fails on the stale path in a connecting/deleting state.

shuo-wu: > one round can still read the namespace device while the next round fails on the stale path in a connecting/deleting state

Will there be a log for this case? A load failure right after a success

derekbit:
Ah @shuo-wu I mistakenly deleted your comment.. Orz I've addressed your comment in server.go and is currently validating it.

pkg/spdk/engine.go
```diff
@@ -377,6 +372,14 @@ func (e *Engine) createNVMeTCPTarget(spdkClient *spdkclient.Client, superiorPort
 	return nil
 }
 
+func (e *Engine) startExposeNVMeTCPTargetLocked(spdkClient *spdkclient.Client, anaState NvmeTCPANAState, spdkANAState spdktypes.NvmfSubsystemListenerAnaState, cntlid uint16) error {
```
shuo-wu: NIT: Why is it called `Locked`?

shuo-wu (commented):
In general LGTM

pkg/spdk/engine.go
derekbit: Removed. Thank you.

shuo-wu (approved):

derekbit:
@mergify backport v1.13.x v1.12.x

PR #690, opened 2026-09-11
fix(enginefrontend): stop reporting an expansion the frontend never finished

derekbit:

Issue Longhorn/longhorn#13993


derekbit:
Closed. The issue seems related to longhorn/longhorn#13965

cc @mantissahz @davidcheng0922

PR #696, opened 2026-09-13
fix: prevent premature expansion completion during frontend reconnect

derekbit:

Issue Longhorn/longhorn#13965 Longhorn/longhorn#13993


pkg/spdk/enginefrontend.go
```diff
@@ -1193,12 +1207,43 @@ func (ef *EngineFrontend) Expand(ctx context.Context, spdkClient *spdkclient.Cli
...
+					engineName, originalSize)
+				return nil
+			}
+
+			// The online RAID expansion tears down and recreates the RAID bdev, forcing
+			// the initiator controller through NVMe error recovery and reconnect. Wait for
+			// the reconnected controller to reach the live state before reporting the
+			// expansion complete, so the new size is not published on a path that is still
+			// reconnecting.
+			if ef.Frontend == types.FrontendSPDKTCPBlockdev {
```
davidcheng0922, line 1228: NIT: Should we use `frontend` instead? We already grab it under the lock,
and the `if` right above uses it — this one reads the field directly.

derekbit: Good catch! Updated. Thanks.

davidcheng0922 (approved):
LGTM

derekbit:
@mergify backport v1.13.x

derekbit:
@mergify backport v1.12.x

PR #699, opened 2026-09-14
fix: should use frontend variable in Expand()

derekbit:

Issue #

Longhorn/longhorn#13965
Longhorn/longhorn#13993


mantissahz (approved):
LGTM
This means ef.FrontEnd might change during the checks in this method, right?

derekbit:
@mergify backport v1.13.x

PR #703, opened 2026-09-15
fix(backup): synchronize snapshot readers with cleanup

christophersherman:

Issue longhorn/longhorn#14017

When a backup source replica stops, backupstore can close the snapshot after the first worker error while other readers are still unwinding. `CloseSnapshot` clears `b.replica`, but `ReadSnapshot` accesses that pointer without the backup mutex. A late reader can panic the shared instance manager and interrupt unrelated replicas.

Capture the replica pointer under the backup mutex, release it before checking replica state, and check the snapshot file under the backup mutex before reading. Readers racing cleanup return an error. The backup and replica mutexes are never held together in `ReadSnapshot`.

Based on current master `1ec706b`. The same bug is present in the v1.12.0 source baseline; please consider a v1.12.x backport.

No running image has been updated. A live detach-during-backup test and the privileged SPDK integration suite (`make ci`) have not been run for this PR.

Validation at `ea5ff67342fc8bcd7d39b64c1238bb8eddfa5baf`, using Go 1.26.0 on Linux/amd64:

- The new post-cleanup regression reproduces the original nil-pointer panic at `backup.go:279` with the parent revision's `backup.go`.
- `gofmt` and `go vet ./pkg/spdk`: pass.
- `go test ./pkg/spdk -count=1`: pass.
- `go test -race ./pkg/spdk -count=3`: pass.
- Linux/amd64 test-binary compilation: pass.

The five tests cover normal reads and EOF, a stopped replica, a missing snapshot file, reads after cleanup, and 32 concurrent readers racing cleanup. Successful reads are checked against the original file contents.

CI: https://github.com/christophersherman/longhorn-spdk-engine/actions/runs/35005385589


pkg/spdk/backup.go
```diff
@@ -276,10 +276,20 @@ func (b *Backup) CompareSnapshot(snapshotName, compareSnapshotName, volumeName s
...
-	b.replica.RLock()
-	replicaState := b.replica.State
-	replicaName := b.replica.Name
-	b.replica.RUnlock()
+	// Cleanup can release the replica while backupstore readers are still
+	// unwinding. Keep a local reference, and drop the backup lock before taking
+	// the replica lock so the two locks are never held together here.
+	b.Lock()
+	replica := b.replica
+	b.Unlock()
```
derekbit, line 284: Good catch!

derekbit (approved):
Thanks @christophersherman for your contribution!

derekbit:
@mergify backport v1.13.x v1.12.x

